PR #2, opened 2017-10-01
Search elpi files in $COQPATH

vbgl:
Plugins and third-party libraries, as coq-elpi,
can be installed in a different prefix than the
standard coq library.

The COQPATH environment variable list search paths
for such libraries.

This patch extends the search path of elpi with the
directories that are listed in this variable.


gares:
Thanks

PR #33, opened 2018-10-11
[coq] Fix Ltac qualification.

ejgallego:
`Ltac_plugin` is a packed module and should be referred in that way.

This works now because `coq_makefile` passes `-I` for packed modules
but this is not compatible with packing per-se I think.


ejgallego:
Included in #34.

PR #45, opened 2018-12-19
Adapt to coq/coq#8817 (SProp)

SkySkimmer:
EDIT: after a rebase of sprop now it works (https://gitlab.com/coq/coq/-/jobs/136854724), no clue what changed but OK.

OUTDATED:
This doesn't actually work but I can't tell why. In test_API.v 
~~~coq

Elpi Query "
 do! [
   coq.env.begin-module-type ""TA"",
     coq.env.add-const ""z"" hole {{nat}} _ _,
     coq.env.add-const ""i"" hole {{Type}} _ _,
   coq.env.end-module-type MP_TA,
   coq.env.begin-module ""A"" MP_TA,
     coq.env.add-const ""x"" {{3}} hole _ _,
       coq.env.begin-module ""B"" _NoGivenModType,
         coq.env.add-const ""y"" {{3}} hole _ GRy,
       coq.env.end-module _,
     coq.env.add-const ""z"" GRy hole _ _,
     coq.env.add-indt (inductive ""i1"" 0 {{Type}} i\ []) I,
     coq.env.add-const ""i"" I hole _ _, % silly limitation in Coq
   coq.env.end-module MP,
   coq.env.module MP L
   %coq.env.module-type MP_TA [TAz,TAi] % @name is broken wrt =, don't use it!
 ]
".
(* Error:
Syntax error: 
near: …uctive "i1" 0 {{Type}} i\ []) I,      coq.env.add-const "i" I hole _ _, % silly limitation in Coq   ….
*)
~~~
Any ideas?

### DIFF
--- a/src/coq_elpi_HOAS.ml
+++ b/src/coq_elpi_HOAS.ml
@@ -58,6 +58,8 @@ let in_coq_name ~depth t =
   | (E.UVar _ | E.AppUVar _) -> Name.Anonymous
   | _ -> err Pp.(str"Not a name: " ++ str (P.Raw.show_term t))
 
+let in_coq_annot ~depth t = Context.make_annot (in_coq_name ~depth t) Sorts.Relevant
+
 (* universes *)
 let univin, isuniv, univout, univ =
   let open CD in
@@ -76,6 +78,7 @@ let univin, isuniv, univout, univ =
   } in
   cin, isc, cout, univ
 ;;
+let sprop  = E.Constants.from_string "sprop"
 let prop   = E.Constants.from_string "prop"
 let typc   = E.Constants.from_stringc "typ"
 let sortc  = E.Constants.from_stringc "sort"
@@ -83,6 +86,7 @@ let in_elpi_sort s =
   E.mkApp
     sortc
     (match s with
+    | Sorts.SProp -> sprop
     | Sorts.Prop -> prop
     | Sorts.Set ->
         E.mkApp typc (E.mkCData (univin Univ.type0_univ)) []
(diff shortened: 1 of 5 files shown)


src/coq_elpi_HOAS.ml
```diff
@@ -76,13 +78,15 @@ let univin, isuniv, univout, univ =
   } in
   cin, isc, cout, univ
 ;;
+let sprop  = E.Constants.from_string "sprop"
```
gares: This change should be also in coq-hoas.elpi

gares:
Also I don't see the point of adding sprop instead of nyi if the binder annotations are hard coded to relevant.  I understand supporting annotations is more work, and I can take care of it, but I need to know if it is necessary

SkySkimmer:
>Sorry, what is the status quo? It works right?

Yes.

>Also, I need to know more about the binder annotations and their behaviour wrt conversion. Is the annotation relevant? I mean is the id function with a relevant annotations convertible with the I'd function with the irrelevant one

The annotations are used in conversion so that `fun x y => x` and `fun x y => y` are convertible when the annotation is Irrelevant but that's all they do.

>Also I don't see the point of adding sprop instead of nyi if the binder annotations are hard coded to relevant.

The annotations get fixed by the kernel, see sprop cep.

gares:
> The annotations are used in conversion so that `fun x y => x` and `fun x y => y` are convertible when the annotation is Irrelevant but that's all they do.

This is not my exact question. An maybe the case I've in mind may never happen, but imagine I pass to the conversion test `Lam ("x",Relevant,Rel 1)  === Lam ("y",Irrelevant,Rel 1)` what does it say?

In elpi `"x"` and `"y"` are identical (the `=` sign in elpi considers them the same). Indeed the conversion test of Coq skips these `Name.t`. What about the relevance annotation (in Coq)? Because I wan the `=` sign of elpi to behave consistently to it.

SkySkimmer:
>what does it say?

It says they're convertible.

gares:
thanks!

ppedrot:
Please merge.

PR #63, opened 2019-06-04
Adapt to Coq PR #8988: change of type internal syntax

herbelin:
This is to be merged synchronously with coq/coq#8988.

Main change is a reversal of the representation of a sort: an anonymous universe is no more encoded as a `Type []` but as an explicit `UAnonymous`, while the named universes are represented using a `UNamed` constructor.


SkySkimmer:
Wrong target branch

herbelin:
> Wrong target branch

You meant that it should be `coq-master`. Done, thanks.

SkySkimmer:
Do merge

PR #66, opened 2019-06-21
[coq] Overlay for coq/coq#10316

ejgallego:


src/coq_elpi_builtins.ml
```diff
@@ -665,11 +667,12 @@ be distinct).|};
              Evd.check_univ_decl ~poly:false evd UState.default_univ_decl in
           Declare.definition_entry
             ~opaque:(opaque = Given true) ?types:ty ~univs:uctx bo in
-       let dk = Decl_kinds.(Global ImportDefaultBehavior, false, Definition) in
```
gares: Where did poly=false go? In case it is an optional argument, I want it to explicitly say false, since coq-elpi supports no universe polymorphic stuff

ejgallego: Umm let me see, IIRC indeed poly is always mandatory for the reasons you state.

ejgallego: Oh, actually poly is not used by that function, before it was there because of the triple being not the right type.

ejgallego: Whether the entry is poly is stored in the entry itself, the PR indeed tries to remove quite a lot of duplication

src/coq_elpi_builtins.ml
```diff
@@ -665,11 +667,12 @@ be distinct).|};
              Evd.check_univ_decl ~poly:false evd UState.default_univ_decl in
           Declare.definition_entry
```
gares: Are you saying that this line is supposed to pass false? Or is it inherited from the check_univ_decl above?

src/coq_elpi_builtins.ml
gares: OK, it looks great, but please check the code above to see if I need to do anything else in order to ensure poly:false in the end

ejgallego: All `poly` arguments are mandatory so unless you pass `poly:true` entries will be monomorphic.

src/coq_elpi_builtins.ml
ejgallego: Yes it is inherited from the check_univ_decl, after the Coq PR you cannot omit `poly` anymore, it is mandatory in all the places it is needed.

ppedrot:
Please merge.

PR #82, opened 2019-09-18
Readd dune

ejgallego:
2d5b28a deleted the dune file, which was not broken per-se but only
usable from a Coq overlay composed build.

We add a version which should work in both contexts, as the composed
build is important in the overlay toolchain.


gares:
The dune build still does not build the .v files. Could you please add that step as well?

ejgallego:
> The dune build still does not build the .v files. Could you please add that step as well?

Actually the dune build of coq-elpi is only useful for checking compilation, loading the plugin will fail as elpi.cma is not linked statically and Coq doesn't know yet how to use findlib to load plugin dependencies.

There are three choices here:
- wait for next Dune version which will allow to link `elpi.cma` statically,
- wait for Coq to add support for findlib's `Dynload`
- merge the ml part only.

Any strategy is fine to me.

gares:
1 is acceptable, 2 is OK, 3 is not ok to me. If I get it right this PR implements 3.

ejgallego:
> If I get it right this PR implements 3.

Yes.

gares:
How are overlay creators that use dune to compile coq ans th eml part of elpi supposed to compile the .v files ?

ejgallego:
> How are overlay creators that use dune to compile coq ans th eml part of elpi supposed to compile the .v files ?

As of today the composed build only works for the ML part, having it work on the `.v` files will take more time, it is not even clear it is possible to compose .v files properly with the current global `user-contrib` library layout.

gares:
Wrt overlays, Coq's dev/ci script calls make, doesn't it? If so, then I don't get 

> tho not having the dune support for the ML parts is a burden to overlay creators.

I don't get which existing workflow can take advantage of this PR, at its current state at least.

ejgallego:
> I don't get which existing workflow can take advantage of this PR, at its current state at least.

The (experimental) workflow is as follows:
- use the `create_overlay` script, over a clean `_build_ci` dir
- uncomment the `(dirs (:standard _build_ci))` line in `dune`
- now the `check` target will provide a fully composed build of Coq + all supported plugins, a single call to dune will identify all parts that break, and you get a fully working merlin setup etc... so the only part is to commit all the fixes and push [to be automated soon, along with the PR creation]

ejgallego:
> * uncomment the `(dirs (:standard _build_ci))` line in `dune`

This is a bit annoying, but soon we will be able to provide a proper workspace for selective composed builds that doesn't require that hack.

gares:
Ok, but overlays are not just for ml changes. And even then, I may want to build the v files to see if the fix is ok. If the elpi plugin built this way cannot be loaded by Coq...

ejgallego:
> Ok, but overlays are not just for ml changes. And even if they were, one may want to build the v files to see if the fix to the ml code is ok.

Indeed, however the ML setup this way is much nicer, once you have done the ml changes you can run the non-composed build to check the .v files.

All overlays I and I think Gaëtan have produced lately used this method and it is a big improvement.

> If the elpi plugin built using cannot be loaded by Coq, I'm not sure it makes sense to have a dune file in the first place.

Most overlays are just fixing compilation so indeed it does help quite a lot in these cases.

ejgallego:
Hi @gares, this should be ready I think, the `.vo` files are built now.

Regarding linking, the default that `coq-makefile` is using, `-linkall` for `cmxs` generation feels pretty dangerous, any other plugin that loads any of elpi transitive deps [such as `ppx_deriving.runtime`] will produce undefined effects [tho in OCaml 4.10 I believe dynlink will just abort], but there is little we can do here unless we modify Coq upstream.

ejgallego:
Travis failure on the regular build seems due to a wrong CI setup, the pin is done to the wrong repository; thus my commit doesn't exist. I've pushed a commit fixing this, see the commit message tho.

gares:
Linkall is not enough, it links what you pass as an argument, but you don't pass all libraries (eg PPx runtime).

I still don't see the .v being compiled.

ejgallego:
Oh indeed I messed up my testing, sorry; I was testing the patch in a composed build with default dune lang "2.0" where `dune build` does indeed build all the targets [as opposed to installable ones]

Let's see now; the build is a bit messy due to coqdep not outputting `.elpi` dependencies from `.v` files, so these have to be tracked manually; there are many possibilities to fix this.

Anyways I still don't see why having the ML-part only in the tree is a problem; it poses no problem and provides significant advantages when writing overlays.

ejgallego:
The `REPO_SLUG` trick for commit testing doesn't work as the merge commit that travis tests doesn't seem to be located in my repos. I guess I'll do a local pin instead.

ejgallego:
> there are many possibilities to fix this.

A possible choice would be to extend coqdep so it can indeed generate the extra deps; or you could have some preprocessing step too; I dunno. The first idea should work modulo the parsing of coqdep itself, which I'm unsure how easy is to modify in these lines.

theories/dune
```diff
@@ -0,0 +1,57 @@
...
+ (package coq-elpi)
+ (files
+   coq-lib.elpi
+   coq-lib-extra.elpi
+   coq-HOAS.elpi
+   coq-elpi-checker.elpi
+   elpi-command.elpi
+   elpi-tactic.elpi
+
+   (bcongr.elpi as derive/bcongr.elpi)
```
gares: Am I correct that copy_files puts everything at the root and here we put it back in place?
This seems quite weird.

src/dune
```diff
@@ -0,0 +1,38 @@
...
+; ocamlfind-based dependency loading (cf coq/coq#7698 )
+(rule
+ (targets elpi_plugin.cmxs)
+ (action (run %{ocamlopt} -shared -linkall -o %{targets}
+           ; The list here is generated using:
+           ; ocamlfind query -predicates native -r -format "%+a" elpi
+           ;
+           ; I've removed some modules that were doubly-linked by Coq,
+           ; this already looks like a bug in CoqMakefile, OCaml 4.10
+           ; will indeed check if the hypothesis is true.
```
gares: This comment is out of place and vague. coq_makefile has
```
CAMLDONTLINK=unix,str
```
which might be an incomplete list, but is there to avoid exactly that. If you had to remove more libraries, you can just open a bug in Coq.

src/dune
```diff
@@ -0,0 +1,38 @@
...
+ (targets coq_elpi_config.ml)
+ (action
+  (with-stdout-to %{targets}
+   (run ocamlfind query -format "let elpi_dir = \"%d\"" elpi))))
+
+(coq.pp (modules coq_elpi_vernacular_syntax))
+
+; We link elpi.cmxa statically into the cmxs; this will break if some
+; other plugin is linking elpi, so it should be replaced by
+; ocamlfind-based dependency loading (cf coq/coq#7698 )
```
gares: I'd like this link to be mentioned here:
https://dune.readthedocs.io/en/stable/advanced-topics.html#dynamic-loading-of-packages

gares:
It looks much better now. The only part that is really uglier than before is the installation code. Can it be fixed?

For the dependencies, is there a way to patch a rule (instead of just writing it)? I mean I'd like the rule for elpi.vo to *also* do `cat whatever.elpi`. This would declare the dependency on the .elpi file, but should also call coqc with the right options. Maybe you can just write down a special one for `elpi.vo` making it depend on `lib/*.elpi`. It would work for me.

theories/dune
ejgallego: `copy_files` puts stuff in the current dir, then I choose to install in a sub-folder as originally; indeed this could use some tweaking.

src/dune
ejgallego: Yeah the current setup is too messy, I will just improve the Coq dynloader so we don't have to rely on these hacks. The doubly-linked problem may arise for example if you load elpi and coq-analytics I think that both use deriving IIRC.

src/dune
ejgallego: After a second thought this is too messy to implement with Dune + Coq Dune [due to the way .v files depend on plugin code], so I propose we add proper tracking to the plugin loader and remove all these hacks in the dune build.

ejgallego:
> It looks much better now. The only part that is really uglier than before is the installation code. Can it be fixed?

Yes, I'll try to improve it; I'm unhappy with the linking setup tho, as the dependencies are really not properly tracked; I'll add instead support for Coq to dynload deps based on the META of the plugin.

ejgallego:
> For the dependencies, is there a way to patch a rule (instead of just writing it)? I mean I'd like the rule for elpi.vo to _also_ do `cat whatever.elpi`. This would declare the dependency on the .elpi file, but should also call coqc with the right options. Maybe you can just write down a special one for `elpi.vo` making it depend on `lib/*.elpi`. It would work for me.

Yes, we should add something like that; there are several ways to do it, including using the dune plugin API, I'll have a look.

ejgallego:
@alizter

ejgallego:
Updated to something that almost works, some issues to solve, apart from fixing Dune to add the right dep for the new loading method in 8.16.

theories/elpi.v
```diff
@@ -1,8 +1,14 @@
+(* This should be covered by dynlink *)
 From Coq Require Ltac.
-Declare ML Module "coq-elpi.elpi".
+
+(* Legacy can't work here *)
+(* Declare ML Module "elpi_plugin:coq-elpi.plugin". *)
+Declare ML Module "coq-elpi.plugin".
 
 (* Generate coq-bultins.elpi *)
-Elpi Document Builtins.
```
gares, line 5: This functionality is crytical, it can be implemented in another way, but I don't know how

ejgallego: We need to tell the rule coqc is generating that file. This is a similar problem than extraction, I think this is something where DAP could help (a way for coqc to tell Dune which targets is generating) cc @Alizter

We can implement this generically, or we could add a bit of support to dune for elpi. What are the build rules of *.elpi stuff usually? Maybe just updating `coqmod` to support this is OK.

I guess the dependency part is covered by coqdep outputting the deps, however, we also need to keep tract of new targets to account for correct build rules.

ejgallego:
[Marking as draft, this is for our testing for now and not ready for merge]

theories/elpi.v
gares: This "command" is totally ad-hoc. One could write a custom binary which links coq, elpi and coq-elpi, and finally runs the code of this command: the list of builtins is first class, and elpi can print a documentation. I keep the file promoted in the repo.

PR #83, opened 2019-10-04
Adapt to coq/coq#10811 (SProp default on)

SkySkimmer:
With SProp on an additional scheme "foo_sind" is automatically created
for each inductive.


ppedrot:
Please merge.

PR #103, opened 2020-01-06
Record declaration as an argument type

### LINKED ISSUE #98: wish: syntax
`Elpi foobar [ 33 | "ident" | ident | (term) | Record <record-syntax> ] .`

gares:
This code now works
```
Elpi Command tele.
Elpi Accumulate lp:{{
main [indt-decl A] :-
  coq.say A,
  coq.typecheck-indt-decl A ok,
  coq.env.add-indt A _.
}}.
Elpi Typecheck.
Elpi tele  Record foo A : Type := {
    a of A & A : A;
    z (a : A) :>  A -> A;
    x (w := 3) : nat;
  }.
Print foo. About z. (* 
Record foo (A : Type) : Type := Build_foo
  { a : A -> A -> A;  z : A -> A -> A;  x : let w := 3 in nat }

For foo: Argument scope is [type_scope]
For Build_foo: Argument scopes are [type_scope function_scope function_scope
                 nat_scope]
z : forall A : Type, foo A -> A -> A -> A

z is not universe polymorphic
Argument scopes are [type_scope _ _ _]
z is a coercion
z is transparent
Expands to: Constant elpi.tests.test_HOAS.z *)

```
The argument `"Record"` is special and begins a record declaration.
I wrote by hand the code to globalize and the declaration and finally convert it to an indt-decl.
It was easier that I thought, but I don't support (and I won't) field notations, eg `{ op : A -> A -> A; op_comm : a * b = b * a where "w * x" := (op w x); }` since the API for notations is imperative (coq declares them while going, then rolls back the state...).

It was easy enough to use the same grammar entries of Coq, so the SSR notations for binders work.

Fix #98


theories/tests/test_tactic.v
```diff
@@ -155,4 +155,11 @@ Qed.
...
 End T1.
 
+From Coq Require Import ssreflect.
 
+Elpi Command tele.
+Elpi Accumulate lp:{{
+main [_, tel T, str F, tel S] :-
+  coq.say T F S.
+}}.
+Elpi tele  Record : x of _ ; x : (w :nat := 3) nat.
```
gares: CC @CohenCyril here `Record` is just like passing `"Record"` and the `x` after `;` is the `str F` argument.

gares:
It is a bit early, since I got some binding wrong ;-)
I'll fix this tomorrow

gares:
@CohenCyril please tell me if this is enough for parameters

src/coq_elpi_goal_HOAS.ml
```diff
@@ -12,23 +12,64 @@ open Names
...
+        let state, fields = do_fields ~depth state fields in
+        let constructor = match constructor with
+          | None -> Name.Name (Id.of_string ("Build_" ^ Id.to_string name))
+          | Some x -> Name.Name x in
+        state, in_elpi_indtdecl_record (Name.Name name) s constructor fields
+    | Glob_term.GProd(name,_,src,tgt) ->
+        let state, src = gterm2lp ~depth state src in
+        let state, tgt = under_ctx name src None do_params ~depth state tgt in
+        state, in_elpi_indtdecl_parameter name src tgt
+    | _ -> CErrors.user_err Pp.(str "now a record declaration")
```
gares: Let-in parameters should have a nicer error message, also there is a typo

CohenCyril:
looks good to me...

PR #109, opened 2020-01-09
Support for Coq's #[attributes]

gares:
This is a bit abrupt.. but let's you do
```
#[attribute] Elpi foo.
```
as well as
```
Elpi Export foo.
#[attribute] foo.
```
Attributes use a generic syntax for command options, such as `poly`, `local`, ...
TODO: a little "parser" written in elpi

### DIFF
--- a/coq-HOAS.elpi
+++ b/coq-HOAS.elpi
@@ -13,11 +13,14 @@
 % Command and tactic invocation (coq_elpi_vernacular.ml)
 % %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-% Entry point for commands. Eg. "Elpi mycommand foo 3 (f x)." becomes
+% Entry point for commands. Eg. "#[att=true] Elpi mycommand foo 3 (f x)." becomes
 %   main [str "foo", int 3, trm (app[f,x])]
-% The encoding of terms is described below.
+% in a context where
+%   attributes [attribute "att" (leaf "true")]
+% holds. The encoding of terms is described below.
 pred main i:list argument.
 pred usage.
+pred attributes o:list attribute.
 
 % Entry point for tactics. Eg. "elpi mytactic foo 3 (f x)." becomes
 %   solve [str "foo", int 3, trm (app[f,x])] <goals> <new goals>
--- a/Changelog.md
+++ b/Changelog.md
@@ -31,6 +31,9 @@ written without the `@`.
 
 ### HOAS
 
+- Change type of `main` to take as the second argument the list of attributes
+  passed to the command invocation (Coq syntax is for example `#[myflag]`).
+  See the attribute-value data type in `coq-builtin.elpi`.
 - Change context entry `def` to not carry a cache for the normal form
   of the defined term (now cached by a specific `cache` context entry).
   `def` now carries the exact same information of a `let`, as `decl`
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [1.3.0] - 2020-01-10
+## [1.3.0] - UNRELEASED
 
 Switch to Elpi 1.9. The main visible change is that opaque data
 types such as `@constructor`, `@inductive` and `@constant` are now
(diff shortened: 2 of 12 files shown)


gares:
I've mixed feelings about this, so I'll decide what to do with this after I have the little parser.

gares:
The alternative would be to have an argument type for attributes, and say that attributes appended/prepended to the arguments when present.

gares:
Looks nicer than expected. Usage mockup
```
main _ Atts :- 
  parse-attributes Atts [
    att "foo" bool,
    att "bar" (oneof [ "xxx" `-> 1,
                               "yyy" `-> 0 ])
 ] Opts,
Opts => std.do! [
  ...
  if (get-option "foo" tt) ...
  if (get-option "bar" 1) ...
  ...
].
```

CohenCyril:
I am a bit confused, I thought there were going to be elpi attributes for other commands, not attributes for elpi commands...
For example, assuming we have a elpi command called `derive_eq` Is there a PR where we can do
```coq
Inductive foo := bar of bool | baz of nat.
#[deive_eq]
```

gares:
Yes, this is true. I'd like that. But this requires making Coq's commands like Definition be able to understand an attribute (implemented in elpi, or in an ML plugins).

What I do here is that I make elpi programs aware of attributes, that are a generic (in the syntax, not in the semantics) options you can pass to any commands. You can write `#[program]` in front of any Coq command, then command can do something sensible or error out saying that he does not support that option. Given that now you can `Elpi Export command` to make `command` available without the `Elpi` prefix, I though to give this command a regular status, that includes being able to see attributes.

gares:
Eg, you could say `#[option] hb.structure RING ...` in order to tune the behavior of `hb.structure`. In a sense these options could be represented as arguments, but I believe that most of the time you gain nothing by mixing the real argument and these options, so better to let you write them with another syntax.

gares:
Here the list of flags that are recurrent in Coq commands: https://coq.github.io/doc/master/refman/language/gallina-specification-language.html#attributes

Given the syntax is open, you can pass what you like to your own commands

CohenCyril:
> Here the list of flags that are recurrent in Coq commands: https://coq.github.io/doc/master/refman/language/gallina-specification-language.html#attributes
> 
> Given the syntax is open, you can pass what you like to your own commands

I have no clue on how to interpret the syntax this paragraph of the coq documentation... :-/

gares:
You can write `#[foo]` or `#[foo="x"]` or `#[foo(bar)]` or `#[foo,baz,..]` or `#[foo(bar="x")]` or `#[foo(bar,baz="y")]`...
You can look at the Rust refman, the syntax is the same.

coq-HOAS.elpi
```diff
@@ -14,10 +14,10 @@
 % Command and tactic invocation (coq_elpi_vernacular.ml)
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-% Entry point for commands. Eg. "Elpi mycommand foo 3 (f x)." becomes
-%   main [str "foo", int 3, trm (app[f,x])]
+% Entry point for commands. Eg. "#[att=true] Elpi mycommand foo 3 (f x)." becomes
+%   main [str "foo", int 3, trm (app[f,x])] [pr "att" (leaf "true")]
 % The encoding of terms is described below.
-pred main i:list argument.
+pred main i:list argument, i:list attribute.
```
CohenCyril: Excuse my ignorance, but I am not sure I understand... 
- First in `coq-builtin.elpi` the add-hoc constructor is used (not `pr`).
- Then d I get it right that your examples from https://github.com/LPCIC/coq-elpi/pull/109#issuecomment-573070631 get translated in the following way
  - `#[foo]` -> `[attribute "foo" (node [])]`
  - `#[foo="x"]` -> `[attribute "foo" (leaf "x")]` BTW, can it only be a string? If so why the double quotes `"`? I thought ot would be possible to do `#[foo=1]` or `#[foo=(fun x => x)]`... And is `#[foo=bar]` correct and represented by `[attribute "foo" (leaf "bar")]`?
  - `#[foo(bar)]`-> `[attribute "foo" (node [attribute "bar" (node [])]]`
  - `#[foo,baz]` -> `[attribute "foo" (node []), attribute "baz" (node [])]`

EDIT: after looking more intensely, I read `parse-attributes`, which if I understand correctly typechecks and interpret the attributes, which are then made available via clauses. Are `attribute` and `attribute-value` reflecting exactly what Coq's API gives, while `parse-attributes` and the typechecking are @gares add-value using elpi?

gares:
We do something similar in coq as well, there is a parser/validator in attributes.ml.

I used the pair constructor in an early version, now I have a specialized clone.

The absence of a value is represented as leaf "" (questionable, but this is what I did)

gares:
So far the only values are strings, no terms yet

gares:
This PR was completely revamped, it should break no code now

PR #110, opened 2020-01-10
parse a bunch of Coq keywords as strings (fix #107)

### LINKED ISSUE #107: Binding symbol arguments
Can we add parsing for arbitrary symbols so that
`Elpi foo x := a.`
has arguments `[str ":=", str "a"]` ?

gares:
Nothing obvious missing? @CohenCyril

### DIFF
--- a/src/coq_elpi_vernacular_syntax.mlg
+++ b/src/coq_elpi_vernacular_syntax.mlg
@@ -179,10 +179,18 @@ let colon_constr = Pcoq.Entry.create "elpi:colon_constr"
 let any_attribute : Attributes.vernac_flags Attributes.attribute =
   Attributes.make_attribute (fun x -> [],x)
 
+let coq_kwd_or_symbol = Pcoq.Entry.create "elpi:kwd_or_symbol"
+
+let the_kwd = ref ""
+let any_kwd strm =
+  match Stream.peek strm with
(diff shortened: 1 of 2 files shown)


CohenCyril:
I can think of many other symbols `+` (used in `Hint Modes`), `/`, `!` (used in `Arguments`) etc...
However `of` is not a keyword, and why is `-` bound to an identifier?

gares:
I wanted to be able to write `foo -x -y -help` (to distinguish arguments from options) but I guess it makes no more sense after #109 .

For the other symbols, I guess I'll try to be more generic next week, that is have a custom parser that accepts any registered symbol and be done with it.

gares:
wrt `of`, it is a keyword to me ;-)
```
Coq < Definition of := 1.
of is defined

Coq < Require Import ssreflect.
[Loading ML file ssrmatching_plugin.cmxs ... done]
[Loading ML file ssreflect_plugin.cmxs ... done]

Coq < Definition of := 1.
Toplevel input, characters 11-13:
> Definition of := 1.
>            ^^
Error:
Syntax error: [Prim.ident_decl] expected after [def_token] (in [vernac:gallina]).
```

gares:
This PR now parses any Coq keyword or symbol (but for `.`) in command mode.

PR #119, opened 2020-02-25
rename coq.gr->* to coq.gref->*

gares:


gares:
@CohenCyril this will break HB

gares:
@CohenCyril this will break HB

Issue #138, opened 2020-05-06
modernize Elpi derive


gares:
- export `Elpi derive`

PR #141, opened 2020-05-07
Inductive as a command argument

### LINKED ISSUE #144: clarify nuparams in `Inductive` argument
In Coq there is no syntax to distinguish between Uniform and Non-Uniform parameters to Inductives. Coq infers it by their usage in the types of the constructor. This is not orthogonal to implicits, since an implicit status may make a parameter automatically inferred (as Uniform?).

This is too complex for Elpi's "inductive" argument, IMO, and also does not reflect how inductives are represented in Elpi: Uniform parameters *do not* occur as arguments to the inductivetype in the type of the constructors (they are fixed, why should you repeat them?), while Non-Uniform (and indexes) *must* appear (they vary, you must tell Elpi how). Coq-Elpi 1.3 only supports `Record` arguments, but I'd like to support also `Inductive` in 1.4 (to be released in few weeks).

I'm leaning toward imposing this very discipline for inductives used as arguments to Elpi commands.
Eg:
```coq
Elpi mycommand
  Inductive ty UP NuP : idx -> Type :=
  | K1 : ty (NuP * NuP) i -> ty NuP j
  | K2.
```
This also makes type checking of constructor with no given type for the conclusion (K2), it is
sufficient to unify they missing type with `ty NuP _` (in this case the index is not inferrable, while the exact argument NuP is exactly what makes NuP a parameter and not an index, but that is the idea, I hope it is understandable).

This is not ideal, since one cannot just prefix any Coq Inductive declaration with, say, `Elpi derive` since he may had implicits on, or use `@` and pass all arguments. But at the same time this design is very simple. The alternative is to have Elpi call the implicit argument machinery during the interning phase (note that Coq does intern + pretype inductive in a mixed way, so Elpi is already running its own code for inductives, since it has to stop after interning.. but so far this code is small, calling impargs does not not really thrill me).

Before committing to this design/compromise I'd love to hear from @maximedenes / @SkySkimmer  since IIRC they did clean up a little inductive + implicits recently.

gares:
Fix #144 #138


gares:
@CohenCyril I'd like to have a second opinion on the changes in coq-HOAS.elpi
In particular the declaration of an inductive got hopefully improved:
- it now carries implicit status for parameters (uniform an non uniform), and the API taking it (and the new one returning it) set/return the implicit status as well. (The point is that Elpi Command Inductive ... can now be almost loseless wrt the info the user can attach to the inductive).
- similarly context-item and const-decl carry more info (even if the env writing API does not understand that yet)

This should round up a little the API we use in HB for builders.

gares:
You may also have comments on example_usage_derive.v and its new `derive Inductive ...` form, that copies some ideas from HB (mostly a spec, I've to finish the coding)

CohenCyril:
Looks good to me.
Howver, shouldn't all `forall` carry implicit status?

gares:
In Coq they can, but only in a few places they actually have any effect. I try to keep terms simple and put metadata elsewhere, like the parameter constructor of an arity.

For terms I may at some point add a node carrying metadata, such as the loc, but the node would be optionally wrapping all subterms. I don't want to make the term data type more complex by default. Moreover metadata makes = wrong, semantically, and I don't like that to be the default.

Thanks for the comments. Next week I'll finish and port HB.

PR #151, opened 2020-07-01
Adapt to coq/coq#12611

ejgallego:
Hi @gares , in the context of coq/coq#12611 , I was reworking the record declaration API and I see that Coq Elpi does use some internals.

I was wondering if you still would like to access that API, or if Elpi could use the higher-level API already in `Record`, which implements the vernacular command.

It seems to me that the current code in Elpi has some overlap with the internals; how would the proper API look for you? Do you think that the route this PR is going would work (at the cost of using the AST), or would you like to go some other route?

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -1030,7 +1030,7 @@ Supported attributes:
      | Some field_specs -> (* record: projection... *)
          let names, flags =
            List.(split (map (fun { name; is_coercion; is_canonical } -> name,
-               { Record.pf_subclass = is_coercion ; pf_canonical = is_canonical })
+               { Record.Internal.pf_subclass = is_coercion ; pf_canonical = is_canonical })
              field_specs)) in
          let is_implicit = List.map (fun _ -> []) names in
          let cstr = (ind,1) in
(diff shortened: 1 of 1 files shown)


gares:
Most of the API in elpi are close to vernac entry points but MUST work on econstr, not glob nor expr. This is why I sometimes call or copy some internal code. I suspect any extension language is in the same position.

gares:
More precisely I d like the elaboration phase from surface syntax to econstr to be separate, at the API level, from the steps that follow, like adding the inductive, synthesizing projections e co. And if possible also have API for the expr->glob and glob->econstr. IIRC, for inductives, they are mixed and Elpi only wants the former (binder analysis, notation expansion). In practice I had to reimplement it in Elpi, since the coq code is entangled. This should give you the full picture.

gares:
Actually glob is not using de Bruijn, but it already knows if x is a global name or a bound one. This is what I call binder analysis.

ejgallego:
I see, thanks for the feedback; for now I think I will keep the internal entry point for Elpi, IMHO we can improve the API quite a bit and do a proper phase-separation later, but we need to be able to get incremental refactorings first.

The case for records is indeed tricky API-wise.

I agree that it would be nice to be more systematic about objects that can be elaborated, etc... including having a generic interface for them; however this kind of change may take a long time to get into Coq proper.

ejgallego:
Ok, for now I've just re-exposed the API, we will see how things turn out later on.

SkySkimmer:
Please merge now

PR #163, opened 2020-07-27
Locality attribute

gares:


Changelog.md
```diff
@@ -15,6 +15,31 @@
 ### API
 - New `coq.arguments.set-default-implicit` that behaves like
   `Arguments foo : default implicits`
+- Change of arguments of type `@global?` attributes `@local!` or `@global!`.
+  In order to pass a locality directive one has to do something like
+    `@global! => coq.add-something`
```
CohenCyril, line 20: Is the semantics of `@global! => (@local! => stuff)` well defined?

CohenCyril: (I think I remember you telling me that `@global!` would win, but I find it counterintuitive)

gares: Hum, local should win.

gares: I'll add a test

gares: What may be unclear to a user is `[ @local! , @global! ] =>foo1` (local wins), but it looks artificial

PR #177, opened 2020-08-21
[app] Namespace emulation

gares:
This new application of Coq-Elpi is motivated by some use cases in HB, but can be of general interest.

TODO:
- [x] move the shared elpi code outside the `theories/namespace.v` file, to `namespace/ns.elpi`, be sure it is installed and that it can be loaded by third parties. Eg by HB in order to implement `HB.mixin Record Namespace.Thing := { ....}`
- [x] move the tests to `theories/namespace/test_namespace.v`
- [ ] document the code and in particular the caveat with the DB (or find a better way to do things)
- [x] think more about the syntax. `NS.begin` and `NS.end` are more HB like

CC @CohenCyril @pi8027


gares:
While this code is promising, I don't think this PR should block the 1.6 release, which enables HB and derive to use the elaborator of Coq (insert coercions in record declarations). Hence I removed it from the 1.6 milestone, which I'd like to close today.

CohenCyril:
Oh the ml part will be added anyway ? Then I agree !

gares:
To me the main issue with the current code is that the way elpi clauses are "stored" does not fit well this use case. I summon @maximedenes since he can probably answer what I propose is doable in Libobject, or not, or a bad idea.

Context: today an Elpi Db can be extended with clauses. These clauses are Libobject "objects" and are hence loaded into the DB following this discipline:
- if the clauses are `Local` then they are dropped at module `End`
- if not, the clauses are loaded only when the module is imported

The main problem in the scenario of this PR is that namespace emulation uses modules, and the clauses we talk about represent the state of the namespaces. So they are kind of meta, but given they are attached to modules (which we fiddle with), we have to go trough some gimmicks in order to have them survive but not duplicate...
The discipline that seems more appropriate for this use case is the one of a superglobal Libobject object, which is horrible for CIC related stuff, bet here we are meta, so it may not be so bad.

My question is:
- is there a way to implement a *single* "object" that has 3 "locality" flags: SuperGlobal, Global, Local? (as of today the object only has the last two)

gares:
Any preferences? I hope to merge this today.

gares:
I also have `NES` for namespace emulation system, but you need to be as old as me to get the joke

gares:
@CohenCyril I found a bug, see usage_NES.v

apps/NES/examples/usage_NES.v
```diff
@@ -0,0 +1,40 @@
...
+   and avoid name collisions.
+   Namespaces are unlinke modules, since you can always add a notion to
+   a namespace, even if the namespace was ended before. *)
+
+NES.Begin This.Is.A.Long.Namespace.
+  Definition stuff := 1.
+NES.End This.Is.A.Long.Namespace.
+
+NES.Begin This.Is.A.Long.Namespace.
+  Definition more_stuff := This.Is.A.Long.Namespace.stuff.
```
gares, line 13: This is also weird, I guess I expect the namespace to be automatically opened.

apps/NES/examples/usage_NES.v
```diff
@@ -0,0 +1,40 @@
...
+(* For convenience one can open a namespace to write short names *)
+NES.Status.
+(*
+NES: current namespace []
+NES: registered namespaces 
+[ns [This] «elpi.apps.NES.examples.usage_NES.This_aux_1.This», 
+ ns [This] «elpi.apps.NES.examples.usage_NES.This_aux_2.This», 
+ ns [This] «elpi.apps.NES.examples.usage_NES.This_aux_3.This»]
+*)
+NES.Open This.Is.A.Long.Namespace. (* BUG: does nothing *)
```
gares, line 39: This is the bug

apps/NES/examples/usage_NES.v
CohenCyril: Ahah true!!

apps/NES/examples/usage_NES.v
CohenCyril: I do not understand... what is the bug?

apps/NES/examples/usage_NES.v
```diff
@@ -0,0 +1,40 @@
...
+Print stuff. (* BUG: = 2 *)
```
CohenCyril, line 40: Is there a `Fail` missing?

gares: It should not fail, that is the bug.

gares: I don't know if there are too few `ns` clauses... or if we should repeat the findall after each module import.

gares:
To me it looks weird that `NES.Begin This.Is.A.Long.Namespace. .. NES.End` only adds clauses like `ns [This] path...` and not `ns [This, Is] path...`

CohenCyril:
> To me it looks weird that `NES.Begin This.Is.A.Long.Namespace. .. NES.End` only adds clauses like `ns [This] path...` and not `ns [This, Is] path...`, `ns [This, Is, A] path...` ...

I think it could be, but that's not how we implemented it...

gares:
Anyway, given it's broken I'm not merging it today. I'll let you fix the bug first. Then I'll try to document the code a bit better.

apps/NES/theories/NES.v
```diff
@@ -60,14 +44,8 @@ Elpi Accumulate Db NES.db.
...
-main [str NS] :- std.do! [
-  std.rev {string->ns NS} Path,
-  std.map {std.findall (open-ns X_ P_)} open-ns->string Stack,
-  std.assert! (std.appendR Path Bottom Stack) "NES: Ending a namespace that is not begun",
-  iter-> Bottom Path end-ns,
-].
-
-main _ :- coq.error "usage: NES.End [DotSeparatedPath]".
+  main [str NS] :- nes.end-path {nes.string->ns NS}.
+  main _ :- coq.error "usage: NES.End [DotSeparatedPath]".
```
CohenCyril, line 48:
```suggestion
  main _ :- coq.error "usage: NES.End <DotSeparatedPath>".
```

apps/NES/theories/NES.v
```diff
@@ -78,14 +56,9 @@ Elpi Command NES.Open.
...
-  string->ns NS RevPath,
-  std.rev RevPath Path,
-  std.map {std.findall (ns Path M_)} ns->modpath Mods,
-  std.spy! (std.forall Mods coq.env.import-module)
-].
-
-main _ :- coq.error "usage: NES.Open DotSeparatedPath".
+
+  main [str NS] :- nes.open-path {nes.string->ns NS}.
+  main _ :- coq.error "usage: NES.Open DotSeparatedPath".
```
CohenCyril, line 61:
```suggestion
  main _ :- coq.error "usage: NES.Open <DotSeparatedPath>".
```

apps/NES/theories/NES.v
```diff
@@ -18,7 +18,7 @@ Elpi Accumulate File "elpi/nes.elpi".
 Elpi Accumulate lp:{{
 
 main _ :-
-  std.map {std.findall (open-ns X_ P_)} open-ns->string Stack,
+  std.map {std.findall (open-ns X_ P_)} nes.open-ns->string Stack,
   coq.say "NES: current namespace" {std.rev Stack},
```
CohenCyril, line 22:
```suggestion
  coq.say "NES: current namespace" {join "." {std.rev Stack}},
```

apps/NES/theories/NES.v
```diff
@@ -0,0 +1,65 @@
...
+}}.
+
+Elpi Command NES.Status.
+Elpi Accumulate Db NES.db.
+Elpi Accumulate File "elpi/nes.elpi".
+Elpi Accumulate lp:{{
+
+main _ :-
+  std.map {std.findall (open-ns X_ P_)} nes.open-ns->string Stack,
+  coq.say "NES: current namespace" {nes.join "." {std.rev Stack}},
```
gares, line 22:
```suggestion
  coq.say "NES: current namespace" {nes.join "." {std.rev Stack} },
```

apps/NES/examples/usage_NES.v
```diff
@@ -0,0 +1,45 @@
...
+   Namespaces are unlinke modules, since you can always add a notion to
+   a namespace, even if the namespace was ended before. *)
+
+NES.Begin This.Is.A.Long.Namespace.
+  Definition stuff := 1.
+NES.End This.Is.A.Long.Namespace.
+
+NES.Begin This.Is.A.Long.Namespace.
+  Definition more_stuff := stuff. (* stuff in the namespace is visible *)
+NES.End This.Is.A.Long.Namespace.
```
herbelin, line 14: I'm arriving in this PR from `coq/ceps#` and I don't know if comments are welcome or not. Let me try though and 
ask whether a shortcut of the form `Definition This.Is.A.Long.Namespace.more_stuff := stuff.` could be supported for the very same three lines above.

CohenCyril: Yes indeed, with ELPI we can emulate it, if we produce a command, such as
```coq
NES.prefixed Definition This.Is.A.Long.Namespace.more_stuff := stuff.
```

apps/NES/examples/usage_NES.v
```diff
@@ -0,0 +1,45 @@
...
+
+Print This.Is.A.Long.Namespace.stuff. (* = 1 *)
+Eval compute in This.Is.A.Long.Namespace.more_stuff. (* = 1 *)
+
+(* Unlike a module, a namespace can contain two notions with the same name.
+   The latter shadows the former. *)
+
+NES.Begin This.Is.A.Long.Namespace.
+   Definition stuff := 2.
+NES.End This.Is.A.Long.Namespace.
```
herbelin, line 24: I did not see the question answered, so I hope this is not redundant. What says `Print This.Is.A.Long.Namespace.more_stuff` after `stuff` has been shadowed?

apps/NES/examples/usage_NES.v
CohenCyril: A big problem in here is the interaction with the section mechanism, which makes it impossible to open an emulated namespace within a section...

apps/NES/examples/usage_NES.v
CohenCyril: The answer is line 28 : `1`

gares: It will probably expose an implementation detail. The two stuff have different kernel names, since we Crete them inside fresh modules. Then via import + the work of the nametab we give the illusion of shadowing.

herbelin: > The two stuff have different kernel names

Could you tell a bit about the implementation details (or tell where to look at to find them)? I.e. what is generated behind the hood?

CohenCyril: Roughly:
- `NES.Begin X.` = `Module Export X_aux_<n>. Module X. NS.Open X.` (where n is a fresh natural number)
- `NES.End X.` = `End X. End X_aux_<n>.`
- `NES.Open X.` = `Import X_aux_0 ... X_aux_<max n>.`

herbelin: Ok, I think I roughly see what it looks like. Interesting that you succeed to do this emulation. So, indeed, this marries badly with the constraints on sections.

What are the longer terms? Do you plan to write something to describe the model you have in mind?

CohenCyril: I'm not sure what's next. Let us try to meet for CEP https://github.com/coq/ceps/pull/25

PR #180, opened 2020-08-28
[NES][FFI] Optimization

gares:
The test_NES_perf file runs Elpi 1000 times. It takes 32 seconds vs 0.5 seconds running "the same" Coq code. The speed of Elpi does not seem to be the culprit. Some API are way slower than their ML counterpart (this overhead is repeated 2K times for begin-module, 2K times for end-module).

Investigation:
- `UState.demote_global_univs` takes 0.0007s on top of APIs taking 0.00007s (around 3s total I guess)
- Elpi's `compiler` (23s in total)
  - on the initial program
    - assembling: 0.0089
    - optimization:0.0062
  - on the last program (500+ ns clauses/compilation units)
    - assembling:0.0406
    - optimization:0.0033
- Elpi's `runtime` (7s in total, including the 3s API overhead)

### DIFF
--- a/apps/NES/_CoqProject
+++ b/apps/NES/_CoqProject
@@ -9,5 +9,7 @@
 theories/NES.v
 
 tests/test_NES.v
+tests/test_NES_perf.v
+tests/test_NES_perf_optimal.v
 
 examples/usage_NES.v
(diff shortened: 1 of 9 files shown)


gares:
By using the new API for program extension in https://github.com/LPCIC/elpi/pull/85 we can implement a cache for programs and quickly append to them clauses (DB entries).

We are now at 11s.
If one disables the `UState.demote_global_univs` call, we are at 6s, so I guess the focus should be there now.

I believe 10x slower than Coq is OK, in the sense that is the speed factor between a compiled and interpreted language.

gares:
merged

PR #186, opened 2020-09-28
Adapt w.r.t. coq/coq#13106

ppedrot:
This is backwards compatible, but I think we need to chat a bit about the feature itself and whether it should really be removed.


gares (approved):

gares:
One less feature to expose to elpi is good news

ppedrot:
As explained in the upstream PR, this feature is not documented, barely tested and virtually unused (one file in math-classes which can be transparently replaced by the backward mechanism). I think you could just have ignored it from the very beginning...

gares:
I think `Coq_elpi_utils.nYI` is enough to ignore a feature no ;-)

ppedrot:
Since @mattam82 agreed with the removal of the feature, I think this can be merged now already.

PR #190, opened 2020-10-14
Universe polymorphism

gares:


gares:
@CohenCyril this is the approach to universe polymorphism we discussed. It is not a huge breaking change, but it is anyway all over the place. The APIs mimics closely the one of Coq: A `gref` cannot be pulled out of the env without applying to it a universe instance. Here these instances can always be omitted, and Elpi picks a fresh one, but you at least to put an `_` even if you don't care about universe polymorphism.

apps/derive/elpi/eqK.elpi
```diff
@@ -100,16 +100,16 @@ main-constructor Lno Prefix K Kty C :- do! [
   coq.env.add-const Name R RT @opaque! EqK,
 
   % add the clause to the db
-  C = (eqK-db K (global (const EqK)) :- !),
+  C = (pi u\ eqK-db K (global (const EqK) u) :- !),
```
gares, line 103: This is maybe the most relevant change: storing univ-instance in a DB is not really what you want so one has to quantify it so that the clause works for any univ-instance.

apps/derive/elpi/invert.elpi
```diff
@@ -65,18 +66,20 @@ trivial-arity (arity (prod Name Ty In)) Ni (parameter ID explicit Ty Out) :-
   ID is "idx" ^ {term_to_string Ni},
   Ni1 is Ni + 1,
   @pi-decl Name Ty p\ trivial-arity (arity (In p)) Ni1 (Out p).
-trivial-arity (arity T) _ (arity T).
+trivial-arity (arity T) _ (arity T1) :-
+  % refresh univs
+  (pi u\ copy (sort (typ _)) (sort (typ u))) => copy T T1.
```
gares, line 71: Hum this looks like a little bug, also this refresh should be part of the standard API

src/coq_elpi_HOAS.ml
```diff
@@ -77,12 +77,16 @@ type options = {
   hoas_holes : hole_mapping option;
   local : bool option;
   deprecation : Deprecation.t option;
+  poly_ind : (bool * UState.universe_decl * UnivNames.universe_binders) option;
```
gares, line 80: these names are not perfect

src/coq_elpi_HOAS.ml
```diff
@@ -506,19 +555,100 @@ let univ =
   API.Conversion.readback = begin fun ~depth state t ->
     match E.look ~depth t with
     | E.UnifVar (b,args) ->
-       let m = S.get um state in
+       let m = S.get um state in (* TODO: remove ???? *)
```
gares, line 558: TODO

src/coq_elpi_HOAS.ml
```diff
@@ -506,19 +555,100 @@ let univ =
...
          let state, u = new_univ state in
          let state = S.update um state (UM.add b u) in
-         state, u, [ E.mkApp E.Constants.eqc (E.mkUnifVar b ~args state) [E.mkCData (univin u)]]
+         state, u, [ E.mkApp E.Constants.eqc (E.mkUnifVar b ~args state) [univin u]]
        end
-    | _ -> univ_to_be_patched.API.Conversion.readback ~depth state t
-  end
+    | _ ->
+      let state, u, gl = univ_to_be_patched.API.Conversion.readback ~depth state t in
+      (*let state, u = purge_1_algebraic_universe state u in*)
```
gares, line 569: RM

src/coq_elpi_HOAS.ml
```diff
@@ -506,19 +555,100 @@ let univ =
...
+      state, u, gl
+  end;
+  embed = (fun ~depth state u ->
+    (*let state, u = purge_1_algebraic_universe state u in*)
```
gares, line 573: RM

src/coq_elpi_HOAS.ml
```diff
@@ -506,19 +555,100 @@ let univ =
...
+  doc = "Constraint between two universes";
+  pp = (fun fmt _ -> Format.fprintf fmt "<todo>");
+  constructors = [
+    K("lt","",A(univ,A(univ,N)),
+      B (fun u1 u2 -> (Option.get @@ Univ.Universe.level u1,Univ.Lt,Option.get @@ Univ.Universe.level u2)),
+      M (fun ~ok ~ko -> function (l1,Univ.Lt,l2) -> ok (Univ.Universe.make l1) (Univ.Universe.make l2) | _ -> ko ()));
+    K("le","",A(univ,A(univ,N)),
+      B (fun u1 u2 -> (Option.get @@ Univ.Universe.level u1,Univ.Le,Option.get @@ Univ.Universe.level u2)),
+      M (fun ~ok ~ko -> function (l1,Univ.Le,l2) -> ok (Univ.Universe.make l1) (Univ.Universe.make l2) | _ -> ko ()));
+    K("eq","",A(univ,A(univ,N)),
```
gares, line 602: Not sure this is really needed

gares (commented):
Still some TODO

Alizter:
Any progress with this?

gares:
Thanks for the interest!
The main problem is that this PR changes so many APIs and the new ones are clearly worse than the old ones.
I thought about this a bit, I think I found a better approach but I did not try to implement it.
I don't have much time in this month, so don't hold you breath.

Alizter:
I'm interested in support for universe polymorphism in HB by the way. So I am guessing this is one of the things that need to happen first.

gares:
You are absolutely right.

PR #191, opened 2020-10-14
Adapt to Coq  PR #13166 : impargs taken into account in record fields (change of type of AssumExpr)

herbelin:
The constructor `AssumExpr` now takes explicitly a binder list, while before a binder list was pushed on the type as soon as `g_vernac.mlg`.

We consider at this point that it is not supported but for compatibility, it could also be pushed to the type using `mkLambdaCN`. Tell me what you would prefer.

Note: to merge at the same time as coq/coq#13166

### DIFF
--- a/src/coq_elpi_vernacular.ml
+++ b/src/coq_elpi_vernacular.ml
@@ -154,10 +154,11 @@ let intern_record_decl glob_sign { name; sort; parameters; constructor; fields }
   let arity = intern_global_constr_ty glob_sign_params @@ CAst.make sort in
   let _, fields =
     List.fold_left (fun (gs,acc) -> function
-    | Vernacexpr.AssumExpr ({ CAst.v = name } as fn,x), { Vernacexpr.rf_subclass = inst; rf_priority = pr; rf_notation = nots; rf_canonical = canon } ->
+    | Vernacexpr.AssumExpr ({ CAst.v = name } as fn,bl,x), { Vernacexpr.rf_subclass = inst; rf_priority = pr; rf_notation = nots; rf_canonical = canon } ->
         if nots <> [] then Coq_elpi_utils.nYI "notation in record fields";
         if pr <> None then Coq_elpi_utils.nYI "priority in record fields";
         let atts = { Coq_elpi_HOAS.is_canonical = canon; is_coercion = inst <> Vernacexpr.NoInstance; name } in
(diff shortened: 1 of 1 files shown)


gares (approved):

herbelin:
@gares: I realize that (at least) the ELPI test `test_HOAS.v` relies on having binders before the `:` in records. I don't feel I'm able to change elpi deeply to explicitly support binders. So, I added a `mkCProdN` to preserve compatibility.

PS: moving the `mkCProdN` out of the parsing phase as I did in #13166 is maybe overkill. I did it on the principle of having the abstract syntax be close to the user syntax, even though it is not clear here that it is important (it would become important if we'd want to support `Eval simpl in` in record fields, as we do for `Definition`).

gares:
I can help, but here I need more context. In Elpi "parameters" are not like binders, in the sense that they can carry more info. It is not just about records, it is in general. Eg `Definition foo (x : T) : forall y : T, P` here `x` and `y` are different, "x" is a relevant name (a string) while "y" is only a pp hint, and also x can carry an implicit status while y cannot. Is this detail relevant here? I'm not sure because you seem to touch let-in declarations in records which, IIRC, are not implemented in Elpi.

herbelin:
> I'm not sure because you seem to touch let-in declarations in records which, IIRC, are not implemented in Elpi.

The PR is about fixing let-in declarations but the PR modifies also the assumptions. Formerly, a record field of the form `f (x:nat) : nat` was passed to `record.ml` into the form `f : forall x:nat, nat`. Now, it is passed as a pair of the context `x:nat` and of the conclusion `nat`. So, if the difference of representation matters for Elpi, something different than what I did should be done indeed (but what?).

gares:
Ok, now I see this. The right thing to do is to change the HOAS of records
  https://github.com/LPCIC/coq-elpi/blob/19ac9ad48b554b115e6b875df1cf5197827bef5c/coq-builtin.elpi#L335
the third component should become an "arity"
  https://github.com/LPCIC/coq-elpi/blob/19ac9ad48b554b115e6b875df1cf5197827bef5c/coq-builtin.elpi#L326
But this is too much for a PR like this one. I'll take care of this myself later.
AFAICT you have a green light with the current code.

herbelin:
OK, thanks.

SkySkimmer:
Please merge now

PR #194, opened 2020-11-16
Adapting to Coq PR #13386: new specific constructor Proj_cs for registering canonical projections which are (primitive) projections

herbelin:
The opportunity of this change is discussed in coq/coq#13386 where it currently seems to be preferred.

Note: I left the same TODO as in the `Const_cs` case.

To be merged synchronously with coq/coq#13386.

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -309,13 +309,19 @@ let cs_pattern =
   doc = "Pattern for canonical values";
   pp = (fun fmt -> function
     | Const_cs x -> Format.fprintf fmt "Const_cs %s" "<todo>"
+    | Proj_cs x -> Format.fprintf fmt "Proj_cs %s" "<todo>"
     | Prod_cs -> Format.fprintf fmt "Prod_cs"
     | Sort_cs _ ->  Format.fprintf fmt "Sort_cs"
     | Default_cs -> Format.fprintf fmt "Default_cs");
   constructors = [
     K("cs-gref","",A(gref,N),
-      B (fun x -> Const_cs x),
-      M (fun ~ok ~ko -> function Const_cs x -> ok x | _ -> ko ()));
+      B (function
+          | GlobRef.ConstructRef _ | GlobRef.IndRef _ | GlobRef.VarRef _ as x -> Const_cs x
+          | GlobRef.ConstRef cst as x ->
+          match Recordops.find_primitive_projection cst with
+          | None -> Const_cs x
+          | Some p -> Proj_cs p),
+      M (fun ~ok ~ko -> function Const_cs x -> ok x | Proj_cs p -> ok (GlobRef.ConstRef (Projection.Repr.constant p)) | _ -> ko ()));
     K("cs-prod","",N,
       B Prod_cs,
       M (fun ~ok ~ko -> function Prod_cs -> ok | _ -> ko ()));


gares:
In the code just below Is like a call to nYI for Proj_cs, in the constructors DSL.

gares (changes requested):
Please deal with proj_cs in the constructors DSL, or one may get a misterious error instead of not yet implemented

herbelin:
I tried this, but I get a mismatch between `Names.GlobRef.t` and `Names.Projection.t`:
```
    K("cs-proj","",A(gref,N),
      B (fun x -> Proj_cs x),
      M (fun ~ok ~ko -> function Proj_cs p -> ok p | _ -> ko ()));
```
I also tried
```
    K("cs-proj","",A(gref,N),                                                                                                
      B (fun x -> Proj_cs x),                                                                                                
      M (fun ~ok ~ko -> function Proj_cs p -> ok (GlobRef.ConstRef (Projection.Repr.constant p)) | _ -> ko ()));             
```
Do I need a way to go from ` x -> Proj_cs x` to some ` x -> Proj_cs (proj-out-of-constant x)`?

gares:
In both cases I think you should call `nYI "bla` which throws an exception. I don't have an Elpi equivalent for projection names (I don't have primitive projections yet)

gares:
https://github.com/LPCIC/coq-elpi/blob/1c2a98255a732590f11f7c4477a97831d85d865a/src/coq_elpi_utils.mli#L10

gares:
Also
```
 K("cs-proj","Not implement yet",A(gref,N), 
```
is welcome (that string appears in the doc)

gares:
I will then refine your patch, but if I don't grep a `nYI` I may forget about it

herbelin:
I pushed something, hoping I understood correctly. Please check.

gares:
> Do I need a way to go from ` x -> Proj_cs x` to some ` x -> Proj_cs (proj-out-of-constant x)`?

I'm sorry I must have misread your initial code. If this is still possible (I don't know) to get a constant out of a projection, then
you should patch the cs-gref case
```
    K("cs-gref","",A(gref,N),
      BS (fun x state ->
        let env = get_global_env state in
        if glob_ref_x_is_a_prim_proj_in env then state, Proj_cs (whatever x)
        else state, Const_cs x),
      M (fun ~ok ~ko -> function Const_cs x -> ok x | Proj_cs x -> ok (GlobRef.ConstRef (Projection.Repr.constant p)) | _ -> ko ()));
```

I can push this code myself later today

herbelin:
> If this is still possible (I don't know) to get a constant out of a projection, then you should patch the cs-gref case

OK, I did as you recommended (changing what had to be changed accordingly).

src/coq_elpi_builtins.ml
```diff
@@ -309,13 +309,20 @@ let cs_pattern =
...
     | Prod_cs -> Format.fprintf fmt "Prod_cs"
     | Sort_cs _ ->  Format.fprintf fmt "Sort_cs"
     | Default_cs -> Format.fprintf fmt "Default_cs");
   constructors = [
     K("cs-gref","",A(gref,N),
-      B (fun x -> Const_cs x),
-      M (fun ~ok ~ko -> function Const_cs x -> ok x | _ -> ko ()));
+      B (function
+          | GlobRef.IndRef _ | GlobRef.VarRef _ as x -> Const_cs x
+          | GlobRef.ConstructRef _ -> assert false
```
gares, line 320: Why assert false?

gares (changes requested):
AFAIK these are canonical values, that is petentially terms

gares:
Thanks, this looks good, bit for the assert false which seems unnecessary

src/coq_elpi_builtins.ml
herbelin: That's an error. Forget about it. I changed it.

gares (approved):

PR #211, opened 2021-02-15
Adding nix support

CohenCyril:


gares:
There is the comment about meta.yaml which I don't have. Also, I don't get anything about nix, so either you add a ci job for it or it will break in matter of minutes.

Also a README explaining what are all these files may be welcome.

gares:
@eponier @bgregoir you may want yo revive this pr

PR #212, opened 2021-02-15
Adding NES.Snapshot

CohenCyril:
NES.Snapshot takes a "picture" of the current state of a Namespace,
for later reuse without the namespace commands.

### DIFF
--- a/apps/NES/examples/usage_NES.v
+++ b/apps/NES/examples/usage_NES.v
@@ -43,3 +43,15 @@ Fail Check nat_def.
 Fail Check @default _ : nat.
 (* This behavior requires Libobject to be aware of the role played by
    a module: if it is a namespace some "actions" have to be propagated upward *)
+
+(* NES Snapshot *)
+(* this allows to take a "snapshot" of a namespace at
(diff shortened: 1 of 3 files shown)


gares:
Why not "NES.export" or "NES.export_as_modules"? I don't think "Snapshot" is very clear

CohenCyril:
"Snapshot" has the advantage to make it explicit that it freezes the current contents. But I would be happy with "export"  too.

apps/NES/examples/usage_NES.v
```diff
@@ -44,14 +44,14 @@ Fail Check @default _ : nat.
 (* This behavior requires Libobject to be aware of the role played by
    a module: if it is a namespace some "actions" have to be propagated upward *)
 
-(* NES Snapshot *)
-(* this allows to take a "snapshot" of a namespace at
+(* NES Export *)
+(* this allows to take a "export" of a namespace at
```
CohenCyril, line 48:
```suggestion
(* this allows to take a "snapshot" of a namespace at
```

gares: hem...

PR #222, opened 2021-03-11
Compatibility with coq/coq#13911.

Zimmi48:
Do not merge yet. This is not a backward-compatible change.
Another option that would be backward-compatible would be to disable the warning locally.


gares (approved):

gares:
don't worry, I'll merge when asked to.

PR #226, opened 2021-03-26
accumulated clauses can survive section if they don't mention variables

gares:


CohenCyril:
I guess I should also disable the CI fior Coq 8.11 and 8.12 temporarily...

PR #229, opened 2021-04-12
More reduction machines

gares:
@pi8027 please have a look

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## UNRELEASED
+
+### API
+- New `coq.reduction.lazy.norm`
+- New `coq.reduction.native.norm`
+- New `coq.reduction.native.available?`
+- Rename `coq.reduction.cbv.whd_all` -> `coq.reduction.cbv.norm`
+- Rename `coq.reduction.vm.whd_all` -> `coq.reduction.vm.norm`
+
 ## [1.9.5] - 26-03-2021
 
 Requires Elpi 1.13 and Coq 8.13.
--- a/Makefile
+++ b/Makefile
@@ -77,9 +77,9 @@ install:
 # compile just one file
 theories/%.vo: force
 	@$(MAKE) --no-print-directory -f Makefile.coq $@
-tests/%.vo: force build
+tests/%.vo: force build Makefile.test.coq
 	@$(MAKE) --no-print-directory -f Makefile.test.coq $@
-examples/%.vo: force build
+examples/%.vo: force build Makefile.test.coq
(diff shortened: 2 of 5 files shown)


Changelog.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## UNRELEASED
+
+### API
+- New `coq.reduction.lazy.norm`
+- New `coq.reduction.native.norm`
+- New `coq.reduction.native.available?`
+- Rename `coq.reduction.cbv.whd_all` -> `coq.reduction.cbv.norm`
```
pi8027, line 9: According to the Coq reference manual, the normalization process of `cbv` recursively applies weak-head normal form computation as in `lazy` (I guess the difference is subterms under lambda abstractions). So you probably want to keep `whd_all` here but not so sure.

gares: For lazy, I could find entry points for weak head normal form, and full normal form. I could not for cbv, which seems to be always recursing under binders. So my "deprecation" is really giving the right name to the code which was already there, for cbv (unless I got mistaken in understanding the Coq API I'm calling)

pi8027: OK. The reference manual may have to be fixed then.

pi8027 (approved):
I confirm that `coq.reduction.native.available?` and `coq.reduction.native.norm` work as I expected.

Changelog.md
gares: I don't think that the `lazy.whd_all`  procedure I bind in coq-elpi is available to the Coq user, so the manual is right, since `Eval lazy` or `lazy` do a full normalization. *but* the kernel of Coq uses "the base case of lazy" for type checking, so the API is there and I bind it as `whd_all` since I believe it is useful.

gares: What I was saying is that I was wrong in thinking that the base case of vm and cbv was there, so I bound them with a `.whd_all` suffix, but I was wrong, since they do `.norm`.

pi8027: For me, the problem is that the quoted paragraph seems to apply to both `lazy` and `cbv` but it is not true for the latter, IIUC.

gares: > Once a weak-head normal form is obtained, subterms are recursively reduced using the same strategy.

This means, IMO, that you obtain a full normal form

pi8027: I agree with that point, but my understanding is that there is no "weak-head normal form" computation (that is different from the full normal form) for `cbv`. So, that sentence does not quite reflect the actual normalization procedure.

In any case, the changes this PR proposes are sensible to me.

PR #235, opened 2021-04-23
lenses

gares:
This PR adds a little gadget to derive. CC @gmalecha @jhaag, have a quick look at the test file to see if this is what you expected. 

TODO:
- [ ] hook it in the main derive call, it won't hurt to generate lenses (when possible)
- [ ] update the apps/derive/README doc

FTR, currently coq-elpi has limited support for primitive records, you can declare them, but you have no syntax for primitive projections (not the Coq's user syntax has...). Still the elaborator of Coq is able to turn a match over a primitive record into an equivalent terms that uses primitive projections, so we always generate a match, for both regular and primitive records, and call the elaborator which knows better.


gares:
Btw, I could not find the laws you were talking about. I guess they are all provable by "case" on the record, so it should be easy to generate them as extra fields of the Lens record, or on the side. But which are these laws?

gares:
So in the end I could figure out 4 laws and I synthesize their proofs, see the file lense_laws.v. The only ugly thing that that their names are, well, ugly.

gmalecha:
This is really cool!

gares:
I just realized that my lenses are simple, I'll look into making them "non simple" tomorrow.

gmalecha:
The wacky lens families are here: but I haven't been able to get them to work nicely: https://github.com/bedrocksystems/coq-lens/tree/lens-family

gmalecha:
Just to expand on my previous statement: I failed to come up with a nice way to get elaboration to work with the lens families discussion (http://comonad.com/reader/2012/mirrored-lenses/)

```coq
Record Lens (inner outer : Type -> Type) : Type :=
  { lens_view : forall {a}, outer a -> inner a
  ; lens_set : forall {a b}, inner b -> outer a -> outer b
  ; lens_over : forall {i i'}, (inner i -> inner i') -> outer i -> outer i'
  ; view_set : forall a b (x : inner b) (init : outer a),
      lens_view (lens_set x init) = x
  ; set_view : forall b (x : outer b),
      lens_set (lens_view x) x = x
  ; set_set : forall a b c (x : inner b) (y : inner c) (init : outer a),
      lens_set x (lens_set y init) = lens_set x init
  ; over_over : forall a b c (f : inner a -> inner b) (g : inner b -> inner c) (init : outer a),
      lens_over g (lens_over f init) = lens_over (fun x => g (f x)) init
  }.
```

What is in in the branch linked above (https://github.com/bedrocksystems/coq-lens/blob/lens-family/theories/LensFamily.v#L4-L20 ) was my attempt to implement it using relations to "parse" types, but I couldn't make them nice to use.

Canonical structures might be a solution to this problem, but I don't understand them well enough to do that.

apps/derive/elpi/lens_laws.elpi
```diff
@@ -0,0 +1,89 @@
...
+  coq.env.add-const Name Bo Ty @opaque! _,
+].
+pred law2 i:indt-decl, i:term, i:term, o:term, o:term.
+law2 (parameter ID _ Ty Rest) Lens Ind {{ fun p : lp:Ty => lp:(Bo p) }} {{ forall p : lp:Ty, lp:(BoTy p) }} :- !,
+  @pi-parameter ID Ty p\ law2 (Rest p) {coq.mk-app Lens [p]} {coq.mk-app Ind [p]} (Bo p) (BoTy p).
+law2 _ Lens IT Bo Ty :-
+  BoS = {{ fun r => lp:{{ {coq.build-match {{r}} IT (law2-rty.aux Lens) law2-bo.aux } }} }},
+  TyS = {{ lib:elpi.derive.lens.law2 lp:Lens }},
+  std.assert-ok! (coq.elaborate-ty-skeleton TyS _ Ty) "ty",
+  std.assert-ok! (coq.elaborate-skeleton BoS Ty Bo) "bo".
```
gmalecha, line 39: Is `coq.elaborate-skeleton` taking something like a term with holes and passing it through the elaborator?

gares: yes, I could not find a better name for this kind of terms. Also, it does not modify it in place (it does not assign the holes), it uses it as a "skeleton".

gares:
About usability, I may lean toward type classes, because CS have (barring hacks) 1 key, while TC can have more. For the 4th law, I need 2 lenses to be used as keys (I would attach to that pair of lenses the lemma which shows they can be exchanged).

apps/derive/elpi/lens_laws.elpi
gares: I did not have the time to document this file, I tried to do a better job in the other one (the one synthesizing lenses)

gmalecha:
Is this due to the fact that you are considering the 4-type variant of `Lens` and not the 2-function variant? Without that, I kind of think that only `Lens a a b b` actually makes sense.

gares:
This is not exactly what I meant.

Take the law `law1_lens1 : set l1 x (set l1 y r) = set l1 x`; it talks about one (concrete) lens `l1`. You can either invoke it by hand, or use an "overloaded" lemma name `set_set : forall l : Lens, lens_laws l -> set l x (set l y) = set l x` where the second argument is a TC instance (a record with 3 proofs as fields, the ones I synthesize, and a lens as a parameter). I could then register `Instance foo : lens_laws l1 := { .... }`, and `rewrite set_set` would work (and leave no side subgoals) on a goal which mentions `l1`.

Now, the 4th law talks about 2 lenses, so I think I need a `multi_lens_laws` class to have 2 arguments (and, for now, one field with a proof of law four. Then have a lemma `exchange : forall l1 l2 : Lens, multi_lens_laws l1 l2 -> set l1 x (set l2 y r) = let l2 y (set l1 x r)` and `Instance bar : multi_lens_laws l1 l2 :=  { ... }`.

This is the usability I was talking about, and I think I need TC, since the second class is hard to define/infer with CS, but maybe we misunderstood each other. Here the parameters of the `Lens` type itself don't play a major role here (I guess).

gmalecha:
I see.

gares:
Merged, I've decided to not play with TC for now and just document and cleanup the thing.
Here the current demo, lenses are at the end:
https://github.com/LPCIC/coq-elpi/blob/master/apps/derive/examples/usage.v

PR #236, opened 2021-04-23
coq.elpi.accumulate: global scope

gares:
Also, Db are superglobal, which means they are there even if not imported,
which meas that we coq-elpi is resilient to loading files in the
wrong order.

Fix https://github.com/math-comp/hierarchy-builder/issues/236


gares:
CC @CohenCyril , this PR requires this change to HB
```patch
diff --git a/HB/common/log.elpi b/HB/common/log.elpi
index 9932b7a..eacdcbb 100644
--- a/HB/common/log.elpi
+++ b/HB/common/log.elpi
@@ -20,7 +20,7 @@ arguments.set-implicit GR I :-  std.do! [
 pred env.add-location i:gref.
 env.add-location GR :-
   if (get-option "elpi.loc" Loc) % remove when coq-elpi > 1.9
-     (log.coq.env.accumulate current "hb.db" (clause _ _ (decl-location GR Loc)))
+     (log.coq.env.accumulate library "hb.db" (clause _ _ (decl-location GR Loc)))
      true.
 
 pred env.add-const-noimplicits i:id, i:term, i:term, i:opaque?, o:constant.
```
It happened to work because of a bug, fixed by this PR

gares:
mathcomp builds as well

PR #247, opened 2021-05-21
Overlay for Coq PR 14354

jfehrle:
This will compile, but if elpi has a dirpath, it should provide it in the second line I changed.  Of course, this only matters if a user wants examine the file in the debugger.


jfehrle:
Use #248 instead.

PR #269, opened 2021-07-12
fix paramX on type aliases (#fix 266)

gares:
CC @CohenCyril I tried to understand the old code and clean it up, I hope you can give a look at it.


apps/derive/paramX-lib.elpi
```diff
@@ -0,0 +1,43 @@
+/* coq-elpi: Coq terms as the object language of elpi                        */
+/* license: GNU Lesser General Public License Version 2.1 or later           */
+/* ------------------------------------------------------------------------- */
+
+namespace paramX {
+
```
gares, line 6:
```suggestion
% [mk-trivial-match N FixTy Acc Bo R] generates R
%   fun (args : tys) (recarg : indty) =>    % all the arguments of the fix up to the rec one
%     match recarg as x in ind idxs return fixty' with
%     | K xs => Bo args' (K xs)
%     end
%  where args' and fixty' are obtained by binding with x::idxs and xs (respectively) items in fixty and args
%  which match the indexes of indty and the indexed of (K xs : indty'').
%  FixTy is convertible to forall (args : tys) (recarg : indty), fixty
```

gares: @CohenCyril in particular I need help to understand if this spec makes sense, is understandable, and is what was implemented before (up-to bug).

CohenCyril: Looks good but for the swap between `args' and fixty'` and `fixty and args` in the explanations...

CohenCyril: Also, `indty''` is not explained

PR #286, opened 2021-09-10
Adapt to Coq PR #14711:  universe contexts and universe binders treated together

herbelin:
Universe contexts and universe binders are treated together leading to some simplifications.

To be merged synchronously with coq/coq#14711. Thanks in advance.

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -828,7 +828,6 @@ let warn_deprecated_add_axiom =
 let add_axiom_or_variable api id sigma ty local =
   let used = EConstr.universes_of_constr sigma ty in
   let sigma = Evd.restrict_universe_context sigma used in
-  let ubinders = Evd.universe_binders sigma in
   let uentry = Evd.univ_entry ~poly:false sigma in
   let kind = Decls.Logical in
   let impargs = [] in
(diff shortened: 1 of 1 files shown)


SkySkimmer:
Please merge now

gares:
I cannot really review this one, maybe @SkySkimmer you can have a look for me? I'll merge in the meanwhile.

src/coq_elpi_builtins.ml
```diff
@@ -837,17 +836,11 @@ let add_axiom_or_variable api id sigma ty local =
     err Pp.(str"coq.env.add-const: the type must be ground. Did you forge to call coq.typecheck-indt-decl?");
   let gr, _ =
     if local then begin
-      let uctx =
-        let context_set_of_entry = function
-          | Entries.Polymorphic_entry uctx -> Univ.ContextSet.of_context uctx
-          | Entries.Monomorphic_entry uctx -> uctx in
-        context_set_of_entry uentry in
-      DeclareUctx.declare_universe_context ~poly:false uctx;
```
SkySkimmer, line 845: Not sure why `poly` was always `false` here

gares: that's on me, since Coq-Elpi only supports universe monomorphic stuff

gares: Hum, indeed @herbelin where is the poly flag now?

gares: it may, in the future. So I guess it is fine for now.

gares: The old code was taken from something in vernac/, so it was probably handling also the poly case, but was dead code really.

PR #289, opened 2021-09-19
Adapt to Coq PR #14846: references to inductive types in the type of constructors are directly Ind rather than Rel nodes

herbelin:
For one part, we insert a translation function which emulates the previous behavior.

There is another part, in file `example_fuzzer.v` (written in Elpi) that I don't know how fix though (I don't see there where it matters that the type is an `Ind` rather than a `Rel`).

This is to be merged synchronously with coq/coq#14846.

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -1180,7 +1180,7 @@ It's a fatal error if Name cannot be located.|})),
      let k_impls = List.map hd k_impls in
      let i_impls = Impargs.extract_impargs_data @@ Impargs.implicits_of_global (GlobRef.IndRef i) in
      let i_impls = hd i_impls in
-     state, !: ((mind,indbo), (i_impls,k_impls)), [])),
+     state, !: (fst i, (mind,indbo), (i_impls,k_impls)), [])),
   DocNext);
 
   MLCode(Pred("coq.env.indc",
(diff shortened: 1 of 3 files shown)


gares:
I suspect the problem is around line 2167. You did amend coq->elpi, but not elpi->coq.

src/coq_elpi_builtins.ml
```diff
@@ -1197,8 +1199,10 @@ It's a fatal error if Name cannot be located.|})),
     let mind, indbo as ind = Inductive.lookup_mind_specif env i in
     let lno = mind.mind_nparams in
     let luno = mind.mind_nparams_rec in
+    let ntyps = mind.mind_ntypes in
     let ty = if_keep ty (fun () ->
       Inductive.type_of_constructor (kon,Univ.Instance.empty) ind
+      |> Inductive.abstract_constructor_type_relatively_to_inductive_types_context ntyps (fst i)
```
gares, line 1205: I don't understand this change, it makes the type point to rels which don't exists. The old API was returning a closed term (see also the `COut(closed_ground_term,"Ty",` line above). But maybe I've misunderstood this line?

gares (changes requested):

src/coq_elpi_builtins.ml
herbelin: You're right, nothing needs to be changed here. I got confused. I'm going to fix it.

SkySkimmer:
Please merge now

SkySkimmer:
ping

gares (approved):

PR #290, opened 2021-09-21
Adapt w.r.t. coq/coq#14903.

ppedrot:


SkySkimmer:
Please merge now

src/coq_elpi_builtins.ml
```diff
@@ -1476,13 +1476,15 @@ and the current module|})))),
     Full(global, {|Declares an inductive type.
 Supported attributes:
 - @primitive! (default: false, makes records primitive)|}))),
-  (fun (me, record_info, ind_impls) _ ~depth env _ -> on_global_state "coq.env.add-indt" (fun state ->
+  (fun (me, uctx, record_info, ind_impls) _ ~depth env _ -> on_global_state "coq.env.add-indt" (fun state ->
      let sigma = get_sigma state in
```
gares, line 1480: @ppedrot why didn't you get uctx from this sigma instead?

ppedrot: As far as I know it could be different if some code is called in the middle, better be safe than sorry.

PR #295, opened 2021-10-01
Fix a few typos

eponier:
Other strange things that I don't know how to patch include:
- in "tutorial_coq_elpi_HOAS.html", the  block "Note: the precedence of lambda abstraction [...]" spans on the next excerpt of Elpi code. I find this strange, but I am not sure this is not done on purpose.
- in "tutorial_coq_elpi_command.html", I am not even sure of the intended meaning of the sentence "Since is a Db is accumulated when a program runs the current contents of the Db are used", so I don't know how to correct it (I think adding a comma would help).
- in "tutorial_coq_elpi_command.html", "use the list to postulate" -> I don't understand either. Should it be "populate"? Or is it a meaning of "postulate" that I don't know?
- in "tutorial_elpi_lang.html", in the explanation of the behaviour of input arguments using the example of "sum", I would at least say a word about why the second argument is also marked as input (even just something like "likewise" or "symmetrically")

### DIFF
--- a/coq-builtin.elpi
+++ b/coq-builtin.elpi
@@ -969,7 +969,7 @@ type never simplification_strategy. % Arguments foo : simpl never
 type when list int -> option int ->
           simplification_strategy. % Arguments foo .. / .. ! ..
 type when-nomatch list int -> option int ->
-                  simplification_strategy. % Arguments foo .. / .. ! .. : simpl moatch
+                  simplification_strategy. % Arguments foo .. / .. ! .. : simpl nomatch
 
 % [coq.arguments.simplification GR Strategy] reads the behavior of the
 % simplification tactics. Positions are 0 based. See also the ! and /
--- a/examples/tutorial_coq_elpi_HOAS.v
+++ b/examples/tutorial_coq_elpi_HOAS.v
@@ -109,7 +109,7 @@ Elpi Query lp:{{
   % destruct GR to obtain its constant part C
   GR = const C,
 
-  % constans may have a body, do have a type
+  % constants may have a body, do have a type
   coq.env.const C (some Bo) TyC
 
 }}.
@@ -381,7 +381,7 @@ Elpi Query lp:{{
 (*|
    
 It is quite frequent to put Coq variables in the scope of an Elpi
-unification variable, and this can be done by sinmply writing
+unification variable, and this can be done by simply writing
 `lp:(X a b)` which is a shorhand for `lp:{{ X {{ a }} {{ b }} }}`.
 
 .. warning:: writing `lp:X a b` (without parentheses) would result in a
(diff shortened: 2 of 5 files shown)


coq-builtin.elpi
```diff
@@ -969,7 +969,7 @@ type never simplification_strategy. % Arguments foo : simpl never
 type when list int -> option int ->
           simplification_strategy. % Arguments foo .. / .. ! ..
 type when-nomatch list int -> option int ->
-                  simplification_strategy. % Arguments foo .. / .. ! .. : simpl moatch
+                  simplification_strategy. % Arguments foo .. / .. ! .. : simpl nomatch
```
gares, line 972: This file is generated from src/coq_elpi_builtins.ml

gares (approved):

coq-builtin.elpi
eponier: Oups, I push a patch to the right file

gares:
Thanks for the corrections, but I need you to fix the first hunk before merge.

>     * in "tutorial_coq_elpi_command.html", I am not even sure of the intended meaning of the sentence "Since is a Db is accumulated when a program runs the current contents of the Db are used", so I don't know how to correct it (I think adding a comma would help).

A Db is like a dynamic file. If you accumulate a file, well, you take the contents of that file once and every time you run the program this does not change even if the file on disk changes. While a Db is is file whose most recent contents are used.
If we were in OCaml, it would a module you recompile and link every time you run the program.

>     * in "tutorial_coq_elpi_command.html", "use the list to postulate" -> I don't understand either. Should it be "populate"? Or is it a meaning of "postulate" that I don't know?

The idea is that once you have a list of things you can postulate a nominal constant with `pi x\` for each and every element of the list (and associate to each of these fresh constants the thing via a binary predicate).

>     * in "tutorial_elpi_lang.html", in the explanation of the behaviour of input arguments using the example of "sum", I would at least say a word about why the second argument is also marked as input (even just something like "likewise" or "symmetrically")

ok

eponier:
> A Db is like a dynamic file. If you accumulate a file, well, you take the contents of that file once and every time you run the program this does not change even if the file on disk changes. While a Db is is file whose most recent contents are used. If we were in OCaml, it would a module you recompile and link every time you run the program.
Ok, I'll try to write a clearer sentence.

> > in "tutorial_coq_elpi_command.html", "use the list to postulate" -> I don't understand either. Should it be "populate"? Or is it a meaning of "postulate" that I don't know?
> 
> The idea is that once you have a list of things you can postulate a nominal constant with `pi x\` for each and every element of the list (and associate to each of these fresh constants the thing via a binary predicate).
Oh, I see, so this really was "postulate". Now that I re-read the sentence and with your explanation, it makes sense. I don't think this needs to be changed (I admit this is where I started not to fully grasp everything of the tutorial, so I may have understood that myself with a more careful read).

gares (approved):

gares:
merging, but the missing serapi for 8.14 will prevent doc from being updated

PR #300, opened 2021-10-22
Adapt to removal of first argument of PrivatePolymorphic in Coq #14727

herbelin:
The first argument of PrivatePolymorphic happened to be useless. It is removed as a side product of the reworking of cooking in #14727. To be merged synchronously.


ppedrot:
@gares Please merge.

PR #319, opened 2021-12-18
Fast generation of Sound Equality Tests

gares:
Current status:
- derivations from feqb are the default (std) collection
- old eq stuff is in the legacy collection
- mathcomp specific code moved to https://github.com/gares/coq-elpi-derive-mathcomp


eponier:
We pushed some nix-related things, that are orthogonal to this PR, because this was convenient. These things are basically the contents of https://github.com/LPCIC/coq-elpi/pull/322. We'll have to decide what to do about them before merging.

eponier:
With @bgregoir, we observed that the behaviour in terms of prefix was not consistent between the commands (some seem even buggy). We decided to systematically add an underscore even when given a prefix from the user, but this probably needs to be discussed at some point. It's a minor point, though.

gares:
> We decided to systematically add an underscore even when given a prefix from the user, but this probably needs to be discussed at some point. It's a minor point, though.

ok

gares:
Now that I've merged the PR about nix, this one needs to be rebased. I can take care of it if you are not working on it today.

eponier:
@gares There is something strange happening with CoqIDE and ProofGeneral (apparently, VsCode works better) due to the fact that there are two _CoqProject files : one for elpi and one for apps/derive. Apparently, CoqIDE and ProofGeneral use the _CoqProject file closest to the current file, meaning in our case the one in apps/derive when we edit some file in derive, and ni  particular some file in apps/derive/tests. But only the _CoqProject in elpi explains how to find the files in the subdirectory test/, meaning that most of the `Require Import` of the files in test/ are not valid. The straightforward solution if we want to edit some file is to temporarily delete the inner _CoqProject, but there is probably a better fix.

gares:
I do run vscode from the root for the main code, and from apps/x for x code. It sucks, but works reliably. I expect coqide to work the same. dunno emacs.

eponier:
Indeed, VsCode has the same behaviour as the two others. Actually, there is a problem only with tests, maybe we should add `-R tests    elpi.apps.derive.tests` to the inner _CoqProject (or something similar?).

gares:
YES

eponier:
@gares done
@bgregoir Now you should be able to open the tests files in Proof General without deleting _CoqProject.

eponier:
I investigated the problems with the tactic `eq_correct_on__solver` (why we need to export things).
- We need to export `ssreflect` because otherwise we have the error `Unknown scope delimiting key ssripat` (cf. https://github.com/coq/coq/issues/15471)
- We need to export `ssrbool` for two reasons
  - because otherwise it does not find `andP` (cf. https://github.com/coq/coq/issues/6687)
  - because otherwise it does not know how to apply a lemma with a conclusion of type `reflect`

Apparently, we don't really need to export `eqb_core_defs`. At least, I cannot reproduce any bug by commenting out that part.

gares:
Now I recall why I did not write the code to generate is_vect_witness :-/

gares:
I did rebase against master, we now need Coq 8.15.
I hope the new nix file is OK.

gares:
Hum, nix is not ok, it is using 8.14. I think it is also broken in master.

gares:
No, the problem is not the version of Coq but rather the one of Elpi, the PR on the nix archive was merged yesterday. @CohenCyril  is there a delay? Do we need to update anything else?

gares:
Great! The Ltac is a bit scary thought :-o

gares:
rebased again (I pushed some fixes to master, error locations did bitrot)

I've added a commit which declares the canonical structure for mathcomp. I think I can make a separate package which registers itself in the main one, so to keep the depedency on mathcomp on the second package only. it is a bit OT here, and will require way more polishing, but it is doable! It has always been, I just did not realize it was.

gares:
nice fix

gares:
oops thanks

gares:
@eponier I'm running out of time today, but `eqbOK` (the one that now just composes correct and refl to get a reflect) is broken.
It should use the eqType ast as the other derivations do, and (more importantly) distinguish between tparam and vparam. Indeed it fails on `ord`, since it thinks the parameter is a type.

I'll resume on monday, but if you have nothing better to do you are welcome.

eponier:
I have the impression you're not so far from merging, nice!

I very probably won't have time to look at it before you, though.

Unrelated: why did you put me as an author of https://github.com/gares/elpi-derive-mathcomp? All the code is from you, isn't it?

gares:
Not sure, it says ask JC

gares:
> I have the impression you're not so far from merging, nice!

very close indeed, and make a release of coq-elpi.

one thing that worries me a little is that the new eqb function apparently don't move when you call simpl. I guess it is the bug that if you hide the match under two deltas, then simpl does not find it. It may be a boring problem for users.

eponier:
> Not sure, it says ask JC

If I recall correctly, Benjamin wrote that when he and I tried to write elpi alone (and it didn't work well :grin: ). Then it was copy-pasted several times. I'd find it more logical if both Benjamin and me are authors, or none of use.

eponier:
> one thing that worries me a little is that the new eqb function apparently don't move when you call simpl. I guess it is the bug that if you hide the match under two deltas, then simpl does not find it. It may be a boring problem for users.

Does `cbn` has the same problem?

gares:
I did not try.

I did find the time to fix the elpi files for eqbok/P but I found a few more usability issues which I want to solve. Make it idempotent, so that running derive twice does not blow up, and allowing one to pass definitions too.

gares:
looks good now. Still I've one question, and IIRC you have this in Jasmin:
```coq
Definition alias := list nat.
derive alias.
```
what should this do? (right now it skips feqb, since alias is not an inductive, but it runs param1 for example).

eponier:
> I did not try.

`cbn` seems to work. And it seems you can use `Arguments eqb_body _ _ _ _ _ _ _ _ _ /.` to make `eqb_body` work with`simpl` too. But there is probably a better `Arguments` command, I don't master well this command.

eponier:
> looks good now. Still I've one question, and IIRC you have this in Jasmin:
> 
> ```coq
> Definition alias := list nat.
> derive alias.
> ```
> 
> what should this do? (right now it skips feqb, since alias is not an inductive, but it runs param1 for example).

It would be great if it would generate `alias_beq := list_beq nat_beq` because I had to turn a definition into a fake inductive to make `feqb` work on the example of the paper.

gares:
> And it seems you can use `Arguments eqb_body _ _ _ _ _ _ _ _ _ /.` to make `eqb_body` work with`simpl` too. But there is probably a better `Arguments` command, I don't master well this command.

Looks fine, would you mind to push this directive in eqb_core_defs.v?

eponier:
Done.

eponier:
I played a bit with the latest version and observed two strange behaviors:
- You need to do some imports, otherwise it fails.
   - derive.eqbOK, otherwise derive does not do much
   - derive.param1, otherwise it misses something about `eq` that is defined manually
   - derive.eqbcorrect, otherwise the tactic `eqb_correct_on__solver` fails
- the idempotence of `derive.eqb` does not work (by the way, I didn't understand why eqbcorrect calls eqbcorrect-for to check whether it has already generated something, why the other commands use a dedicated predicate `xxx_done`)

gares:
ill push some fixes late.

the idea is that you load derive.std, or derive.legacy or .....

eponier:
2 bugs fixed, 2 bugs remain:
- derive.param1, otherwise it misses something about eq that is defined manually
- derive.eqbcorrect, otherwise the tactic eqb_correct_on__solver fails

eponier:
```
From elpi.apps Require Import derive.std.

Inductive t : Set :=  C : forall n : nat, Nat.eqb n 5 = true -> t.

#[only(eqbcorrect)] derive t.
(* derive.param1: No unary parametricity translation for @eq *)
```

gares:
what a fight! thanks!

gares:
mostly fixed, go figure how Coq modules work

eponier:
```
Require Import String.
From elpi.apps Require Import derive.std.

Inductive t := | C : forall s : string, String.eqb s "dffd"%string = true -> t.

#[verbose] derive Ascii.ascii.
#[only(eqb,param1)] derive string.
#[only(param1_trivial)] derive bool.
#[verbose,only(param1_inhab)] derive Ascii.ascii.
#[only(param1_inhab)] derive string.
#[only(param1)] derive Bool.eqb.
Module Import M.
#[only(param1)] derive Ascii.eqb.
End M.
Module Import M2.
derive String.eqb.
End M2.
#[only(eqbcorrect)] derive Top.t.
```
shows
```
derive: no topological order: [] 
[dep1 eqbcorrect-alias eqb-alias, dep1 eqbOK-alias eqbcorrect-alias]
```

eponier:
```
Require Import String.
From elpi.apps Require Import derive.std.

Inductive t := | C : forall s : string, String.eqb s "dffd"%string = true -> t.

#[verbose] derive Ascii.ascii.
#[verbose,only(param1,eqType_ast)] derive string.
(* #[only(param1_trivial)] derive bool. *)
(* #[verbose,only(param1_inhab)] derive Ascii.ascii. *)
#[only(param1_inhab)] derive string.
#[only(param1)] derive Bool.eqb.
Module Import M.
#[only(param1)] derive Ascii.eqb.
End M.
Module Import M2.
#[only(param1)] derive String.eqb.
End M2.
#[only(eqbOK)] derive Top.t.
```
shows
```
??: eqb-for (global (indt «string»)) (global (indt «string»)) 
 (X0^2 c4 c5 c6 c7 c8 c9)
```
This is comprehensible, but not clean.

eponier:
```
From elpi.apps Require Import derive.std.

Inductive bla := D.
Inductive t := | C : bla -> t.

#[only(eqType_ast,param1,param1_inhab)] derive bla.
derive t.
```
shows
```
derive.eqbcorrect: boolean equality for t, maybe use derive.eqb first
```
But I expected `derive` to generate `eqb` itself. Probably it should fail earlier at `eqb` with a nice error message.

eponier:
```
From elpi.apps Require Import derive.std.

Inductive bla := D.
Inductive t := | C : bla -> t.

Elpi derive.eqType.ast bla.
Elpi derive.eqType.ast t.
Elpi derive.param1 bla.
Elpi derive.param1 t.
Elpi derive.induction t.
Elpi derive.tag t.
Elpi derive.fields t.
Elpi derive.tag bla.
Elpi derive.fields bla.
Elpi derive.eqb bla.
Elpi derive.eqb t.
Elpi derive.param1.trivial is_bla.
Elpi derive.param1.trivial is_t.
#[only(induction)] derive bla.
Elpi derive.eqbcorrect t.
```
fails with
```
The elpi tactic/command derive.eqbcorrect failed without giving a specific error message. Please report
this inconvenience to the authors of the program.
```
I cannot reproduce this with only `derive` calls, but maybe due to the topological bug.

gares:
Thanks for all this testing. I think I solved most of the bugs.
I'm still working on a avoid-name-clash patch which I hope to finish tomorrow.

gares:
@eponier I believe all known bugs are somehow fixed, if you have some more time to test it, it would be welcome.

eponier:
Thanks, I'll give a look soon!

eponier:
I pushed a commit where I use fresh names in param1 and param2. There are other files to adapt too I think. But at least now, `#[recursive] derive value` (where `value` is the Jasmin type) works. I also chose to ignore the prefix in param1 and param2, because it seems to me that it does not make a lot of sense to have things called `nat_is_nat`; `is_nat` is already clear.

apps/derive/elpi/eqbcorrect.elpi
```diff
@@ -0,0 +1,441 @@
...
+  param1-functor-for ISGR Funct BitMask, !,
+  apply-functor (global Funct) BitMask L Rec TOTO,
+  apply-aux Aux L Aux1, 
+  R = {{ fun (x : lp:GRL) H => lp:Aux1 x (lp:TOTO x H) }},
+].
+% no params, no aux lemma (no reali argument)
+search What _Rec (global GR as GRL) {{ fun (x : lp:GRL) (_ : lp:IsGR x) => lp:R x }} :-
+  What (global GR) R,
+  mk-reali (global GR) IsGR, !.
+search What _ X _ :- coq.safe-dest-app X HD _, std.assert! (What HD _) "run eqbcorrect before.".
```
eponier, line 416:
```suggestion
search What _ X _ :- coq.safe-dest-app X HD _, std.assert! (What HD _) "run eqbcorrect before".
```

apps/derive/elpi/derive.elpi
```diff
@@ -4,82 +4,118 @@
...
+  coq.env.dependencies T _ AllDeps,
+  coq.gref.set.elements AllDeps AllDepsL,
+  std.filter AllDepsL indt-or-const Deps,
+  main.aux Deps [] CL1,
+  CL1 => main1 T P CL2,
+  std.append CL1 CL2 CL.
+pred main.aux i:list gref, i:list prop, o:list prop.
+main.aux [] X X.
+main.aux [T|TS] Acc CL :-
+  /* (pi X\get-option "only" X :- !, fail) => */ Acc => main T tt CL1,
```
eponier, line 103: Is it done on purpose, or is it some leftover?

apps/derive/elpi/derive.elpi
```diff
@@ -4,82 +4,118 @@
...
 namespace derive {
 
+pred exists-indc i:inductive, i:(constructor -> prop).
+exists-indc I P :-
+  coq.env.indt I _ _ _ _ KL _,
+  std.exists KL P.
+
 pred if-verbose i:prop.
-if-verbose P :- get-option "verbose" tt, !, P.
+if-verbose P :- (get-option "verbose" tt ; get-option "recursive" tt), !, P.
```
eponier, line 13: Is it done on purpose, or was it introduced for debugging reasons?

apps/derive/examples/usage.v
```diff
@@ -77,3 +67,25 @@ Check Box._tag_set_set : (* a Lens law *)
 Check Box._tag_contents_exchange : (* another one *)
   forall A (r : Box A) x y, set Box._tag x (set Box._contents y r) =
                             set Box._contents y (set Box._tag x r).
+
+(** Finally, one can derive an existing inductive typegenerated constants are
+    prefixed with nat_ but won't be in the right
+    place, which is where the type is defined. This means that two users
+    may run derive for the same type in different files, leading to
+    duplication. *)
```
eponier, line 75: The first sentence is unclear, a "." is missing?

coq-builtin.elpi
```diff
@@ -775,6 +780,11 @@ type coq.inline.no coq.inline. % Coq's [no inline] (aka !)
 type coq.inline.default coq.inline. % The default, can be omitted
 type coq.inline.at int -> coq.inline. % Coq's [inline at <num>]
 
+% [coq.env.fresh-global-id The ID one wants The fresh id closer to ID,
+% starting from 1] Generates an id which is not taken (in the current
+% module)
```
eponier, line 785: I know this is automatically generated, but this is not really readable.

eponier (commented):
I read the code and have a few comments.

eponier:
Still valid:
```
Require Import String.
From elpi.apps Require Import derive.std.

Inductive t := | C : forall s : string, String.eqb s "dffd"%string = true -> t.

#[verbose] derive Ascii.ascii.
#[verbose,only(param1,eqType_ast)] derive string.
(* #[only(param1_trivial)] derive bool. *)
(* #[verbose,only(param1_inhab)] derive Ascii.ascii. *)
#[only(param1_inhab)] derive string.
#[only(param1)] derive Bool.eqb.
Module Import M.
#[only(param1)] derive Ascii.eqb.
End M.
Module Import M2.
#[only(param1)] derive String.eqb.
End M2.
#[only(eqbOK)] derive Top.t.
```
shows
```
??: eqb-for (global (indt «string»)) (global (indt «string»)) 
 (X0^2 c4 c5 c6 c7 c8 c9)
```
This is comprehensible, but not clean.

eponier:
Still valid:
```
From elpi.apps Require Import derive.std.

Inductive bla := D.
Inductive t := | C : bla -> t.

#[only(eqType_ast,param1,param1_inhab)] derive bla.
derive t.
```
shows
```
derive.eqbcorrect: boolean equality for t, maybe use derive.eqb first
```
But I expected `derive` to generate `eqb` itself. Probably it should fail earlier at `eqb` with a nice error message.

eponier:
All the other bugs are fixed, thanks!

But I discovered one more that is really strange. `#[only(param2)] derive Ascii.ascii.` fails with the error `No binary parametricity translation for Set` but `Elpi derive.param2 Ascii.ascii` succeeds.

coq-builtin.elpi
gares: oops, dunno what went wrong

gares: fixed

apps/derive/elpi/derive.elpi
gares: I did it on purpose, since I believe `recursive` is a sort of debugging option.

apps/derive/elpi/derive.elpi
gares: This was the thing I was explaining the other day during lunch, that recursively I was enabling all derivations.
It seems to work even if I don't do it, so I guess I was wrong and the commented code can go.

gares:
This time I think I've fixed them all.

apps/derive/elpi/derive.elpi
eponier: Hum, if I try
```
Inductive t := | C : forall n, Nat.eqb n 3 = true -> t.

#[only(eqbOK),recursive] derive t.
```
I get
```
Derivation eqbOK unknown or not applicable to input const «Nat.eqb»
```
so probably it was useful to override the options passed to derive.

gares: fixed

gares:
@eponier that should fo it, I got confused with lists and rev :-/

eponier:
@gares This is nice! Here is a list of the minor things that remain:
- `trivial` calls `inhab`: if I'm not mistaken, the entanglement comes from `is_eq_witness` which requires `trivial` on the type parameter, so indeed you cannot call `inhab` separately from `trivial` in the tests, but I'm not sure that it means that we have to mix the codes (btw, does it make sense to have two rules associated to `is_eq` in `param1-inhab-db`?)
- "witness" vs "inhab": I would call everything "inhab" for the sake of uniformity
- clash of names: here and there, especially for constructors as I mentioned

None of this is blocking, and this PR becomes too large, so if you want you can merge. The bugs can be fixed as they are triggered.

gares:
I think renaming the witness lemmas is a blocker, all the rest can be improved later

eponier:
Nice! :tada:

PR #324, opened 2022-01-12
[HOAS] 1 slot cache for context readback

### LINKED ISSUE #320: Context readback can be a performance bottleneck
... as I reported in math-comp/apery#4. I use the following versions (revisions) of packages for benchmarking:
- OCaml 4.12.1+flambda+fp
- memtrace.0.2.1.2
- Coq 8.15+rc1 + a memtrace patch [(commit)](https://github.com/pi8027/coq/tree/8.15+memtrace)
- coq-bignums.8.15.0
- coq-mathcomp-(ssreflect|fingroup|algebra|solvable|field).1.13.0
- coq-mathcomp-finmap.1.5.1
- coq-mathcomp-bigenough.1.0.1
- coq-mathcomp-real-closed.1.1.2
- coq-mathcomp-multinomials.dev [(commit)](https://github.com/math-comp/multinomials/commit/ff9303bbac534d0cfaf4075deff053abde30f9a0)
- coq-paramcoq v8.15 branch [(commit)](https://github.com/coq-community/paramcoq/commit/fc86b66b3d46deb2658593cd3ab6dd1f30fb28d0)
- coq-coqeal.1.1.0
- coq-elpi.1.12.0
- coq-mathcomp-zify.1.1.0+1.12+8.13
- coq-mathcomp-algebra-tactics.0.2.0
- coq-mathcomp-apery [(commit)](https://github.com/pi8027/apery/commit/59976f118792c5eb43dbfbe523235973393ca533)

The trace file, obtained by the following command, is available on [Gist](https://gist.github.com/pi8027/e2d28a69241f423b0e54b445fc08d768). One may use [memtrace-viewer](https://github.com/janestreet/memtrace_viewer) to visualize it.
```
MEMTRACE=trace.ctf make -j theories/ops_for_b.vo 
```

@gares

gares:
CC @pi8027 , please bench on algebra tactics

I'll sleep on cache invalidation, which is the hard stuff, but the results I have are worth all the thinking!
Thanks for pinning down this issue.

Fix #320

### DIFF
--- a/src/coq_elpi_HOAS.ml
+++ b/src/coq_elpi_HOAS.ml
@@ -2433,11 +2433,23 @@ under_coq2elpi_relctx ~calldepth state params
 (* ********************************* }}} ********************************** *)
 (* ****************************** API ********************************** *)
 
+let ctx_cache = Ephemeron.K1.create ()
+
 let get_current_env_sigma ~depth hyps constraints state =
 (* TODO: cahe longer env in coq_engine for later reuse, use == on hyps+depth? *)
   let state, _, changed, gl1 = elpi_solution_to_coq_solution constraints state in
+  if changed then Ephemeron.K1.unset_data ctx_cache;
   let state, coq_ctx, gl2 =
-    of_elpi_ctx ~calldepth:depth constraints depth (preprocess_context (fun _ -> true) (E.of_hyps hyps)) state (mk_coq_context ~options:(get_options ~depth hyps state) state) in
+    match Ephemeron.K1.(get_key ctx_cache, get_data ctx_cache) with
+    | Some k, Some (c,e) when k == hyps && not changed && e == Global.env () ->
+        (*Printf.eprintf "hit\n";*)
+        state, c, []
+    | _ -> 
+        (* Printf.eprintf "miss\n";*)
+        of_elpi_ctx ~calldepth:depth constraints depth (preprocess_context (fun _ -> true) (E.of_hyps hyps)) state (mk_coq_context ~options:(get_options ~depth hyps state) state)
+  in
+  Ephemeron.K1.set_key ctx_cache hyps;
+  Ephemeron.K1.set_data ctx_cache (coq_ctx,Global.env ());
   state, coq_ctx, get_sigma state, gl1 @ gl2
 ;;
 let get_global_env_current_sigma ~depth hyps constraints state =
--- a/_CoqProject.test
+++ b/_CoqProject.test
@@ -32,3 +32,4 @@ tests/test_cache_async.v
 tests/test_COQ_ELPI_ATTRIBUTES.v
 tests/perf_calls.v
 tests/test_require_bad_order.v
+tests/test_ctx_cache.v
--- a/tests/test_ctx_cache.v
+++ b/tests/test_ctx_cache.v
@@ -1,0 +1,47 @@
+From elpi Require Import elpi.
+
+Elpi Tactic perf.
+Elpi Accumulate lp:{{
+
+pred loop i:int, i:prop.
+loop 0 _.
+loop M P :-
+  N is M - 1,
+  P,
+  loop N P.
+
+solve (goal _ _ _ _ [str "cache", int N]) _ :- !,
+  loop N (coq.unify-eq {{ 0 + 0 }} {{ 0 }} ok, @pi-decl `x` {{ bool }} x\ true).
+solve (goal _ _ _ _ [str "nocache", int N]) _ :- !,
+  loop N (@pi-decl `x` {{ bool }} x\ coq.unify-eq {{ 0 + 0 }} {{ 0 }} ok, true).
+
+
+}}.
+Elpi Typecheck.
+
+Notation t := 
+    (
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat *
+        nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat * nat    
+    )%type.
+Time Goal
+forall (x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 : t),
+forall (x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 : t),
+forall (x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 : t),
+forall (x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 : t),
+forall (x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 : t),
+True.
+intros.
+Optimize Heap.
+Time elpi perf nocache 3000.
+Optimize Heap.
+Time elpi perf cache   3000.
+
+trivial.
+Qed.


pi8027:
Before (coq-elpi.1.12.0):
```
COQC theories/ops_for_b.v
Reification: 6.187799 sec.
Reflection: 85.895381 sec.
Tactic call ran for 92.421 secs (92.229u,0.195s) (success)
Reification: 0.666850 sec.
Reflection: 0.184950 sec.
Reification: 0.002756 sec.
Reflection: 0.009220 sec.
Reification: 4.881952 sec.
Reflection: 2.024934 sec.
Tactic call ran for 7.095 secs (7.081u,0.011s) (success)
Reification: 116.987483 sec.
Reflection: 23.108200 sec.
Tactic call ran for 141.648 secs (140.96u,0.76s) (success)
theories/ops_for_b.vo (real: 391.64, user: 390.39, sys: 1.31, mem: 10241860 ko)
```
After (this branch):
```
COQC theories/ops_for_b.v
Reification: 5.748859 sec.
Reflection: 85.116045 sec.
Tactic call ran for 91.204 secs (91.081u,0.148s) (success)
Reification: 0.659214 sec.
Reflection: 0.179425 sec.
Reification: 0.002202 sec.
Reflection: 0.009155 sec.
Reification: 4.375008 sec.
Reflection: 2.186131 sec.
Tactic call ran for 6.771 secs (6.781u,0.004s) (success)
Reification: 101.330215 sec.
Reflection: 21.836223 sec.
Tactic call ran for 124.906 secs (124.152u,0.76s) (success)
theories/ops_for_b.vo (real: 374.65, user: 373.41, sys: 1.28, mem: 9927092 ko)
```

pi8027:
So it seems this is rather a minor improvement at least for now. Another thing in my mind is caching structure instances that passed conversion checking, to reduce the number of calls to `coq.unify-eq`. But, it seems to make my reification code more complicated.

gares:
I was thinking about the sanders example witout clearing the 3 huge context entries.

In apery it is indeed minor, still visible tough.

gares:
the code has commented printf for hit/miss, you may want to check that the ratio is good in your case.

pi8027:
In the large example (`Lemma from_sander`) of math-comp/algebra-tactics#41, the hit/miss ratio is 11/32754.

gares:
Not sure, because there is just one slot.
Maybe I do something silly elsewhere, please re-profile.

pi8027:
The result of profiling `coqc -R theories mathcomp.algebra_tactics examples/from_sander.v` (both memtrace and perf) is available here: https://gist.github.com/pi8027/eea729a82da78204dd6b49a5d8453852

It seems to me that `lp2constr` still allocates a significant amount of objects, and the overhead of [`maux`](https://github.com/LPCIC/elpi/blob/ad3a854c4f137e4bb132e3d967eb3ea5dd95e280/src/runtime.ml#L881) (18.37%) is quite large (although I don't understand what it does).

gares:
OK, I'll try to run it locally and dig deeper.
`maux` is moving a term from one context to another, depending on the input it may lift (as in shifting De Bruijn levels) or check for out of scope names and de-lift (like the clear tactic in some sense). It is impossible to guess when/why it is called. I need to dig.

pi8027:
I see. Thanks a lot.

gares:
I pushed two commits. I did measure with the opam switch you suggested, but I did not profile memory (lack of time). On my laptop:
```
---- vanilla
COQC theories/ops_for_b.v
Reification: 4.751434 sec.
Reflection: 57.010342 sec.
Reification: 0.442901 sec.
Reflection: 0.153301 sec.
Reification: 0.002216 sec.
Reflection: 0.008304 sec.
Reification: 3.779674 sec.
Reflection: 1.592206 sec.
Reification: 97.784707 sec.
Reflection: 18.144977 sec.

---- 1 cache
COQC theories/ops_for_b.v
Reification: 4.157101 sec.
Reflection: 55.832140 sec.
Reification: 0.429373 sec.
Reflection: 0.145292 sec.
Reification: 0.001683 sec.
Reflection: 0.019609 sec.
Reification: 2.988997 sec.
Reflection: 1.510377 sec.
Reification: 79.990848 sec.
Reflection: 17.583797 sec.

---- 1 cache + gref
COQC theories/ops_for_b.v
Reification: 3.697429 sec.
Reflection: 57.605122 sec.
Reification: 0.469188 sec.
Reflection: 0.166291 sec.
Reification: 0.001312 sec.
Reflection: 0.008518 sec.
Reification: 2.673028 sec.
Reflection: 1.565326 sec.
Reification: 60.440112 sec.
Reflection: 17.521636 sec.

----- 1 cache + gref + ephemeron cache
COQC theories/ops_for_b.v
Reification: 3.796424 sec.
Reflection: 56.384071 sec.
Reification: 0.418464 sec.
Reflection: 0.141059 sec.
Reification: 0.001217 sec.
Reflection: 0.019024 sec.
Reification: 2.340404 sec.
Reflection: 1.380822 sec.
Reification: 55.012264 sec.
Reflection: 16.701333 sec.

----- 1 cache + gref + hashtbl cache
Reification: 3.508264 sec.
Reflection: 56.365453 sec.
Reification: 0.417178 sec.
Reflection: 0.151394 sec.
Reification: 0.001152 sec.
Reflection: 0.007071 sec.
Reification: 2.314913 sec.
Reflection: 1.373354 sec.
Reification: 52.184148 sec.
Reflection: 16.719829 sec.
```

gares:
If you have the time, I'd be happy to look at the memtrace data.

pi8027:
Thanks. This is really great! I updated [the Memtrace data](https://gist.github.com/pi8027/e2d28a69241f423b0e54b445fc08d768)  of `ops_for_b.v` and [the Perf and Memtrace data](https://gist.github.com/pi8027/eea729a82da78204dd6b49a5d8453852) of `Lemma from_sander` of math-comp/algebra-tactics#41.

pi8027:
Note that large proof contexts are still a significant slowdown factor, e.g., [this change](https://github.com/math-comp/apery/pull/4/commits/1ec59edd0992ffa779be4a51e7620346ac998a1b) still saves 30 seconds. But, clearing some items does not always lead to a visible performance improvement (this I didn't notice until now). Maybe, proof-local `def`s depending on section variables are particularly problematic? (I will try to produce an example.)

gares:
Hum, I'll look into letins, their encoding may be a problem. pi-def loads two things, one contains a hole to store the normal form, but that may disable the cache ...

gares:
The data is *very* interesting.
There is quite some allocations coming from `Elpi.Constraints.Ice.freeze`. 
That code is:
- naive (uses List.map instead of smart_map, so it reallocates a term even if it did not change)
- should not be called often, I guess it is called to purge an evar from the constraint store (and the context this evar lives in is huge).
- the CHR rule which fires ignores the context, so there is no need to freeze it (freezing means replacing evars with constants, so that unification behaves like matching in some sense).
- purging one constraint (of the form `evar Ctx Evar Type`) could even be a primitive, no need to implement it via a CHR rule as I do here: https://github.com/LPCIC/coq-elpi/blob/dabdf40259c5f51b8f64c91cecac652db7931d23/elpi/coq-HOAS.elpi#L165

So there is a lot of room for improvement (but I need to patch and release elpi).

It also explains why this has something to do with the context being small.
It is a bit unclear to me why this matters so much, I suspect your code passes to coq a term with many holes (with the huge context) and they are solved, so I fire a lot of time the CHR rule which then freezes the same context over and over. To be confirmed.

pi8027:
> It also explains why this has something to do with the context being small. It is a bit unclear to me why this matters so much, I suspect your code passes to coq a term with many holes (with the huge context) and they are solved, so I fire a lot of time the CHR rule which then freezes the same context over and over. To be confirmed.

I'm lost. I think that terms to reify do not have any evar.

pi8027:
Maybe, the recognition of morphisms is the source of this issue? https://github.com/math-comp/algebra-tactics/blob/8176f8aecdc7288d88ba743bf1f6055ba661370c/theories/common.elpi#L372-L374

gares:
Could be, but I think it is a legit code. It is just that elpi is too naive (for now).

pi8027:
As a side note, this rule is called for every variable (in terms of reification) in the input term and then fails, which is bad. So I prefer to eventually replace this rule with ones auto-generated from canonical instances of morphisms. Nevertheless, I think we will not be able to get rid of this kind of evars. So, it is indeed great if you can fix that.

gares:
The patch on the elpi side to make CHR rule faster was trivial, but it did not change much the overall perf.
The root problem seems to be that whenever I create an evar on the elpi side I do make a copy of the current context (and run a goal like `HugeCtx => declare-evar Ev Ty Ev`. Half of the time is spent in copying that huge context from Coq to Elpi, in particular let-ins, which have a large body in this proof (and not only to read it back, which is now cached). I have think more how to improve on that.

src/coq_elpi_HOAS.ml
```diff
@@ -990,15 +1119,41 @@ and in_elpi_evar ~calldepth k state =
...
+      let state, rest = generate_actual_goals state rest in
+      state, API.Conversion.RawGoal (E.mkAppL rm_evarc [raw_ev; ev]) :: rest
+  | (API.Conversion.Unify _ | API.Conversion.RawGoal _) as x :: xs ->
+      let state, xs = generate_actual_goals state xs in
+      state, x :: xs
+  | _ -> assert false
+
+
+let () = E.set_extra_goals_postprocessing (fun l state ->
+  generate_actual_goals state l
```
pi8027, line 1155:
```
# File "src/coq_elpi_HOAS.ml", line 1155, characters 2-23:
# 1155 |   generate_actual_goals state l
#          ^^^^^^^^^^^^^^^^^^^^^
# Error: This function has type
#          Elpi__API.State.t ->
#          API.Conversion.extra_goal list ->
#          S.t * API.Conversion.extra_goal list
#        It is applied to too many arguments; maybe you forgot a `;'.
```

gares:
Sorry, an incomplete push. I've also pushed another optimization. Now I have:
```
Reification: 0.602927 sec.
Reflection: 0.155692 sec.
Reification: 0.000693 sec.
Reflection: 0.005916 sec.
Reification: 0.981206 sec.
Reflection: 1.254999 sec.
Reification: 16.618525 sec.
Reflection: 15.127710 sec. (* the 2 extra seconds *)
Finished transaction in 32.723 secs (32.391u,0.328s) (successful)
Finished transaction in 4.121 secs (4.084u,0.035s) (successful)
```
The 2 extra seconds were taken to move the call ltac, accounted for in the maux column. It is still high, but now smaller than vmcompute. I think I'll stop optimizing and consolidate all this code and release all that.

gares:
According to my measurements:
- Time: 97+18 -> 16+15
- Memory: 10G? -> 2.5G

Memory is a bit still high...

pi8027:
It seems that my `field` tactic is now slightly faster than `rat_field` which is really great, but reflection (`vm_compute`) seems to become slower in Coq 8.15 (compared to 8.14) which I don't understand why... I don't see any major concerns regarding the paper, but it would be nice if you can backport this PR to Coq-Elpi for Coq 8.14. Again, thanks a lot for the great improvements!

gares:
if so, open an issue on coq.
i wont release for 8.14, sorry no time right now.
for the end of the week I hope to have it out on 8.15

pi8027:
> but reflection (vm_compute?) seems to become slower in Coq 8.15 (compared to 8.14) which I don't understand why...

Sorry, this was not true. I'm confused since I cannot reproduce a similar performance to the one I measured before. I will sort it out.

README.md
```diff
@@ -55,7 +55,7 @@ At the time of writing Proof General does not handle quotations correctly, see P
 In particular `Elpi Accumulate lp:{{ .... }}.` is used in tutorials to mix Coq and Elpi code
 without escaping. Coq-Elpi also accepts `Elpi Accumulate " .... ".` but strings part of the
 Elpi code needs to be escaped. Finally, for non-tutorial material, one can always put
-the code in an external file and use `Elpi Accumulate File "filename".` instead.
+the code in an external file and use `Elpi Accumulate File "filename" From elpi.apps.derive.` instead.
```
gares, line 58:
```suggestion
the code in an external file and use `Elpi Accumulate File "filename" From some.load.path.` instead.
```

gares:
FYI @pi8027 I did split this PR into smaller one (in case you are benching). I'll review/fix merge them all ASAP.

PR #331, opened 2022-01-24
API: coq.hints.opaque

### LINKED ISSUE #329: [wish] Hint Opaque
It would be useful to support the generation of `Hint Opaque` declarations from within Elpi. E.g., 
```
Definition r : nat := 0.
#[global] Hint Opaque r : some_database. (*<-- API for writing this declaration*)
```

The current API around hints:
```
% -- Coq's Hint DB -------------------------------------

% Hint Mode
kind hint-mode type.
type mode-ground hint-mode. % No Evar
type mode-input hint-mode. % No Head Evar
type mode-output hint-mode. % Anything

% [coq.hints.add-mode GR DB Mode] Adds a mode declaration to DB about
% GR.
% Supported attributes:
% - @local! (default: false)
external pred coq.hints.add-mode i:gref, i:string, i:list hint-mode.

% [coq.hints.modes GR DB Modes] Gets all the mode declarations in DB about
% GR
external pred coq.hints.modes i:gref, i:string, o:list (list hint-mode).
```
doesn't seem to support this.

See also https://coq.zulipchat.com/#narrow/stream/253928-Elpi-users.20.26.20devs/topic/Hint.20Opaque/near/268738505.

gares:
Minimal API to declare opacity hints.

CC @gstew5 , please tell me if the API is sufficient for your use case

Fix #329

### DIFF
--- a/coq-builtin.elpi
+++ b/coq-builtin.elpi
@@ -903,7 +903,8 @@ type mode-output hint-mode. % Anything
 % [coq.hints.add-mode GR DB Mode] Adds a mode declaration to DB about
 % GR.
 % Supported attributes:
-% - @local! (default: false)
+% - @local! (default is export)
+% - @global! (deprecated)
 external pred coq.hints.add-mode i:gref, i:string, i:list hint-mode.
 
 % [coq.hints.modes GR DB Modes] Gets all the mode declarations in DB about
@@ -913,10 +914,11 @@ external pred coq.hints.modes i:gref, i:string, o:list (list hint-mode).
 % [coq.hints.set-opaque C DB Opaque] Like Hint Opaque C : DB (or Hint
 % Transparent, if the boolean is ff).
 % Supported attributes:
-% - @local! (default: false)
+% - @local! (default is export)
+% - @global! (deprecated)
 external pred coq.hints.set-opaque i:constant, i:string, i:bool.
 
-% [coq.hints.opaque C DB Opaque] Reads if constant C is opauqe (tt) or
+% [coq.hints.opaque C DB Opaque] Reads if constant C is opaque (tt) or
 % transparent (ff) in DB
 external pred coq.hints.opaque i:constant, i:string, o:bool.
 
--- a/coq-builtin.elpi
+++ b/coq-builtin.elpi
@@ -910,6 +910,16 @@ external pred coq.hints.add-mode i:gref, i:string, i:list hint-mode.
 % GR
 external pred coq.hints.modes i:gref, i:string, o:list (list hint-mode).
 
+% [coq.hints.set-opaque C DB Opaque] Like Hint Opaque C : DB (or Hint
+% Transparent, if the boolean is ff).
+% Supported attributes:
+% - @local! (default: false)
+external pred coq.hints.set-opaque i:constant, i:string, i:bool.
+
+% [coq.hints.opaque C DB Opaque] Reads if constant C is opauqe (tt) or
+% transparent (ff) in DB
+external pred coq.hints.opaque i:constant, i:string, o:bool.
+
 % -- Coq's notational mechanisms -------------------------------------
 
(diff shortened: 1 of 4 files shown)


coq-builtin.elpi
```diff
@@ -910,6 +910,16 @@ external pred coq.hints.add-mode i:gref, i:string, i:list hint-mode.
 % GR
 external pred coq.hints.modes i:gref, i:string, o:list (list hint-mode).
 
+% [coq.hints.set-opaque C DB Opaque] Like Hint Opaque C : DB (or Hint
+% Transparent, if the boolean is ff).
+% Supported attributes:
+% - @local! (default: false)
+external pred coq.hints.set-opaque i:constant, i:string, i:bool.
+
+% [coq.hints.opaque C DB Opaque] Reads if constant C is opauqe (tt) or
```
gstew5, line 919: Just a typo:
```suggestion
% [coq.hints.opaque C DB Opaque] Reads if constant C is opaque (tt) or
```

gares: Thanks! (this file is generated, so I won't merge this code suggestion)

coq-builtin.elpi
```diff
@@ -910,6 +910,16 @@ external pred coq.hints.add-mode i:gref, i:string, i:list hint-mode.
 % GR
 external pred coq.hints.modes i:gref, i:string, o:list (list hint-mode).
 
+% [coq.hints.set-opaque C DB Opaque] Like Hint Opaque C : DB (or Hint
+% Transparent, if the boolean is ff).
+% Supported attributes:
+% - @local! (default: false)
```
Blaisorblade, line 916: I don't think we need this right now (@gstew5 can correct me if I'm wrong), but wouldn't you eventually want to support `#[export]` locality as well for all hints? Would supporting that break the API?

src/coq_elpi_builtins.ml
```diff
@@ -1966,19 +1983,48 @@ Supported attributes:
...
+     Hints.add_hints ~locality [db] Hints.(HintsTransparencyEntry(HintsReferences [r],transparent));
+     state, (), []
+    ))),
+  DocAbove);
+
+  MLCode(Pred("coq.hints.opaque",
+    In(constant, "C",
+    In(hint_db, "DB",
+    Out(B.bool, "Opaque",
+    Easy {|Reads if constant C is opauqe (tt) or transparent (ff) in DB|}))),
```
Blaisorblade, line 2020: Typo, in the autogeneration source:
```suggestion
    Easy {|Reads if constant C is opaque (tt) or transparent (ff) in DB|}))),
```

coq-builtin.elpi
Blaisorblade: Actually, IIUC this line of code:
```
let locality = if options.local = Some true then Hints.Local else Hints.Export in
```

We're likely to need `#[global]` as well, at least since sometimes we sometimes use modules as namespaces.

More in general, `#[export] Hint Opaque` is reasonable but not as a default: a reasonable proof might spawn typeclass searches for modules that are required but not imported, but that shouldn't forget their opacity settings.

gares: My understanding was that export was the (future) default, and global was deprecated. I can add global if you need it, but if I got it right I won't make it the default. @ppedrot

Blaisorblade: "Not default" sounds very acceptable :-).

gares: @ppedrot could you confirm that:
- default: export
- optional: local
- deprecated: global

is future proof?

gares:
There you are, if I hear nothing I'll merge tomorrow.

coq-builtin.elpi
ppedrot: Yes. You can still consider that `global` is not deprecated, it's just never going to be the default anymore.

PR #332, opened 2022-01-25
Add support for functors and funsigs

gstew5:
Add support for generation of
- `Module`s with arguments (Functors);
- `Module Type`s with arguments (Funsigs).

Add two test cases to `test_API.v`.

NOTE: It seems like the code I wrote generates the right functors/funsigs. But it's definitely possible I've done something totally wrong (I've never touched the Coq AST before). 

I may also be missing updates to parts of the codebase I'm unfamiliar with. So far, I've touched only `coq_elpi_builtins.ml` and fixed the resulting downstream breakages.


gares:
I believe the code is good. It will break hierarchy-builder. I wonder if the API should be called coq.env.begin-functor and have coq.env.begin-module be some elpi code passing the empty parameter list.

gares:
I guess you also want an API like `coq.env.apply-module-functor`. Given the name you gave to `binders_ast` I guess you have found it already, if not here is the code you should imitate:
 https://github.com/coq/coq/blob/20b040dd851e9c7a733008dd3035388a7966f885/vernac/vernacentries.ml#L1185

gstew5:
Thanks. I think I've addressed your points about backwards compatibility with HB and the doc string (changed now to `FunctorArguments` in both cases).

I'll work on functor application next.

coq-builtin.elpi
```diff
@@ -685,16 +685,29 @@ external pred coq.env.add-indt i:indt-decl, o:inductive.
...
 
-% [coq.env.begin-module Name ModTyPath] Starts a module, the modtype can be
-% omitted *E*
-external pred coq.env.begin-module i:id, i:option modtypath.
+% [coq.env.begin-module-functor Name ModTyPath FunctorArguments] Starts a
+% module, the modtype can be omitted *E*
+external pred coq.env.begin-module-functor i:id, i:option modtypath, 
+                                           i:list (pair id modtypath).
+
+%
```
gares, line 693: you want to add a `pred ...` line to silence a warning

gstew5: Thanks. The warnings should be fixed now.

PR #334, opened 2022-01-26
Testing reverse dependencies with nix

CohenCyril:


gares:
I've opened another PR, but didn't you foget to regenerate the actions file?

CohenCyril:
> I've opened another PR, but didn't you foget to regenerate the actions file?

sure, I did it via the github interface so I couldn't

gares:
OK, so I'll keep mine and override more when it works

PR #335, opened 2022-01-26
[ci] enable HB

gares:


gares:
@CohenCyril any clue about analysis, which fails with an anomaly?

CohenCyril:
> @CohenCyril any clue about analysis, which fails with an anomaly?

no clue...

PR #339, opened 2022-02-04
Prepare Coq-Elpi 1.13

### LINKED ISSUE #306: Typos
There are typos here and there in the sources, which lintian pointed out and I'm preparing a patch.

### LINKED ISSUE #336: Enrich `Elpi Accumulate File` to search `From` some logical Coq path
Currently coq-elpi uses the [following function](https://github.com/LPCIC/coq-elpi/blob/master/src/coq_elpi_vernacular.ml#L180) in order to generate the valid paths with which to search when invoking `Elpi Accumulate File`. In short, this function results in the following list of search paths: `[./., ./.., ......lib/elpi, ......user-contrib/elpi]`; of note is the fact that `./.` and `./..` are interpreted **relative to the directory in which the coqc process was invoked**.

Not only does this behavior break Emacs Proof General clients who wish to step over such commands - since Proof General (seemingly) starts the coqc process from the directory containing the file being stepped through - but it also breaks CI setups in which the build occurs in a location that differs from that which developers use locally. At least the latter of these issues is a major blocker for use of `Elpi Accumulate File` within my team (and I am personally affected by the former, although many on my team would be unaffected). 

[This Zulip thread](https://coq.zulipchat.com/#narrow/stream/253928-Elpi-users.20.26.20devs/topic/.60Elpi.20Accumulate.20File.60.20and.20Coq's.20.60LoadPath.60/near/270211136) contains the initial question which prompted the creation of this issue; [a subsequent comment](https://coq.zulipchat.com/#narrow/stream/253928-Elpi-users.20.26.20devs/topic/.60Elpi.20Accumulate.20File.60.20and.20Coq's.20.60LoadPath.60/near/270237017) that I made outlined a plausible solution which @gares plans to investigate. The proposed solution involves a new piece of Elpi Vernacular of the form `Elpi Accumulate File FILE From LOGICAL_PATH` and would have the semantics "Accumulate the `.elpi` file `FILE` relative to the physical path referenced by the Coq logical path `LOGICAL_PATH`".

### LINKED ISSUE #337: [wish] Hint Resolve
It would be useful to support the generation of `Hint Resolve` declarations from within Elpi. E.g.,

```
Definition MyVeryCoolWrapper := ....
#[global] Hint Resolve MyVeryCoolWrapper | PRIORITY : my_db my_other_db. (*<-- API for writing this declaration*)
```

The current API around hints does not support this; Enrico believes that - like the request for [Hint Opaque] support in [this issue](https://github.com/LPCIC/coq-elpi/issues/329) - this will be a relatively easy feature to add.

https://coq.zulipchat.com/#narrow/stream/253928-Elpi-users.20.26.20devs/topic/Hint.20Resolve/near/270286873

gares:
Fix #306
Fix #336
Fix #337


gares:
@jhaag please take a look at the commit for `coq.hints.add-resolve` (a better test is probably needed)

@jhaag please take a look at commit [Use Accumulate File ... From loadpath everywhere](https://github.com/LPCIC/coq-elpi/pull/339/commits/a9e608c57eb47b87c65ffd562df71b580e0ba194) for the new resolver mechanism

@pi8027 I've fixed the issue about `coq.option.add` not surviving a file, but I could not find your issue (just don't open it)

pi8027:
That was fast. I confirm this PR fixes the issue about `coq.option.add`.

coq-builtin.elpi
```diff
@@ -934,6 +934,13 @@ external pred coq.hints.set-opaque i:constant, i:string, i:bool.
 % transparent (ff) in DB
 external pred coq.hints.opaque i:constant, i:string, o:bool.
 
+% [coq.hints.add-resolve GR DB Priority Pattern] Like Hint Resolve GR |
+% Priority : DB.
+% Supported attributes:
+% - @local! (default is export)
+% - @global! (discouraged, may become deprecated)
```
jhaag, line 956: @ejgallego I don't see an `@export!` macro in [coq-builtin.elpi](https://github.com/LPCIC/coq-elpi/blob/master/coq-builtin.elpi); do you plan to add this, and how should the `#[export]` attributed be added to without such a macro?

gares: It is the default (if you say nothing, that is what you get). BTW, I'm @gares ;-)

gares:
I have an easy plan for coqdep as well, I'll ping you when I'll submit a PR on Coq (bet before I want to finish this one, last thing missing is functor application which is 80% done; CC @gstew5 I did push here your commit but GH does not display your avatar... git tells me you are still the author tough).

coq-builtin.elpi
jhaag: Haha, of course :upside_down_face:

I had a total brain fart; apologies for the noise

gstew5:
> I did push here your commit but GH does not display your avatar... git tells me you are still the author tough).

Thanks. No worries at all about the avatar.

> last thing missing is functor application which is 80% done

Sorry I didn't get to that. Doing fully general functor application support seemed like it needed new syntax for module expressions, e.g., in terms like `Module X (Y : Z A) := Foo Bar.` I'm curious to see how you approached it.

gares:
I'm planning to only allow simple arguments, eg you would need to name Z A in your example.

Does this sound like a good start? Or it is too simplistic?

gares:
That is also true in you patch btw (if you omit the := ... part).

gstew5:
> I'm planning to only allow simple arguments, eg you would need to name Z A in your example.

This seems pretty reasonable. And I don't think this reduces expressivity (you can always factor out intermediate definitions).  

> That is also true in you patch btw (if you omit the := ... part).

You're right :-)

Blaisorblade:
> > I'm planning to only allow simple arguments, eg you would need to name Z A in your example.
> This seems pretty reasonable. And I don't think this reduces expressivity (you can always factor out intermediate definitions).

What if your functor is dependently typed like "Module (X : A) (Y : B X)."? Not much to factor out. (How important it is, up to you.)

gares:
I think you should be able to "un curry" it by making a first functor with one parameter, then naming B X, then building another functor.

Not nice, but if you don't have a use case right away I'd just stop here and wait for for another feature request to add what is missing.

I have the feeling that modules and functors are a meta programming device (in the logic, brr). Elpi like tools give you another angle at this aspect, so hopefully one needs functor less (but I have no proof of that, so here some functor support).

PR #345, opened 2022-03-09
option to elaborate arguments

gares:
TODO:
- [ ] cleanup
- [ ] give a syntax to the flag
- [ ] write and backport a patch wich ignores all unknown attributes


gares:
@gstew5, FYI.

One would need to be able to select (or not) the raw mode for arguments. The last commit hardwired the new mode of argument processing (for `Definition`). One option I'm thinking at is at Export time, eg
```
#[arguments(raw)] Elpi Export mlock. (* current default *)
#[arguments(elab)] Elpi Export mlock.
```

gstew5:
> `#[arguments(raw)] Elpi Export mlock. (* current default *) ...`

This seems like reasonable syntax. Choosing the encoding at command definition time makes sense.

I wouldn't think it would be useful to choose the encoding at command invocation time, e.g., using an attribute like
```
#[arguments(elab)] mlock Definition ...
```
That might also be harder to implement anyway.

gares:
Yup. In the end I've attached the info to `Elpi Command` (since you can invoke a command even before exporting it, so it felt better). I'm also thinking about changing the default, the only tool out there which really needs raw arguments is HB, and I can patch it myself (but it is also the tool which motivated "coq declarations argument" in the first place, so the default seemed right back then).

gares:
superseded by https://github.com/LPCIC/coq-elpi/pull/360

PR #346, opened 2022-03-10
Fixing the makefile

CohenCyril:


CohenCyril:
> yes, tell me when I can merge

whenever, but the docker CI is failing is it normal?
(I'll come back to fix Nix CI later)

PR #347, opened 2022-03-15
Fix an error in the comment that explains copy

ybertot:
A "." should be replaced by a ":-" for this explanation to make sense (IMHO).


gares (approved):

gares:
Luckily the actual code of `copy` is correct ;-)

PR #348, opened 2022-03-21
Adapt w.r.t. coq/coq#15837.

ppedrot:


gares:
You will need to explain me more if you want to get this merged.
This is how things are now: `sort (typ "u0")` = `Type@{u0}`, and `sort prop` = `Prop` and `sort sprop` = `SProp`.

Now you call `u0` a sort? What would sort polymorphism abstract on? I guess the whole argument of `sort` no?

ppedrot:
See the upstream PR for more details. Not sure what elpi is doing exactly, but morally nobody should use `Universe.t` explicitly. Today, this type is isomorphic to Sorts.t, and all its uses should be turned into the latter instead. Soon a universe will just be an algebraic expression between indices (so no Prop, no Set, just variables, globals and a root), and a sort will be a quality (e.g. Prop, Type, SProp) paired with an algebraic level with some additional quotients. In this view Set := Type[0], Prop := Prop[i] for all i (the quotient is ensured statically), same thing for SProp. As such, a universe won't be enough to know if you're Prop or Set, since they're represented the same for this type.

gares:
So you are saying that `(sort typ u0)` and `(sort prop i0)` and `(sort sprop j0)` are the new sorts after that PR? Or that this is what you aim to?
I'm trying to understand the terminology you want to establish, in the main PR I don't get it either.

In this PR you (partially) rename `u0` "Sort.t" (while now it is a "Universe.t"). From a user perspective, u0 is clearly not a sort, only `sort Kind Index : term` is.

coq-builtin.elpi
```diff
@@ -792,7 +792,7 @@ external pred coq.env.primitive-projections i:inductive,
 
 % -- Universes --------------------------------------------------------
 
-% Univ.Universe.t
+% Sorts.t
 typeabbrev univ (ctype "univ").
```
gares, line 796: This is the thing I don't get, it can't be the right terminology

ppedrot:
> Or that this is what you aim to?

Yes.

> From a user perspective, u0 is clearly not a sort

There is a bit of notational abuse here. A sort is a `Sorts.t`, but by language abuse it is also the term that is built out of it with the `Sort` constructor of `constr`. In any case, basing the user-facing names on the kernel implementation is a very bad idea because the kernel confuses several notions in the same `Universe.t` type, which is what needs to be solved before moving further to a cleaner model.

coq-builtin.elpi
ppedrot: I am fine with whatever you want to call that for the end-user. The only important point is that nobody should rely on the `Universe.t` type in a plugin without a good reason.

ppedrot:
(If that helps you, in the new model `Universe.t` could be renamed `Algebraic.t` or something, i.e. a formal maximum of indices with increments, where indices are just names in a rooted graph.)

gares:
But for the comment referring to the ML module name, a `univ` in elpi is the index of `typ`, a universe level. So just avoid using `Sort.t` as a comment, "universe level" would be fine.

ppedrot:
Not sure I understand what your asking. Note also that in Gallina syntax, `Type@{Prop}` is a thing, so I don't think it is meaningful in the long run to keep a type of universe levels separate.

gares:
The only thing I want is to give syntax to what has one.

I'd like to have users write `prop` in their rules and have that match `Prop` or `Type@{Prop}`, the semantic value, not what the user could write in Coq. Levels don't have a syntax, you can bind them to variable and use APIs for them.

If I get this right, the PR is not changing that, it is only moving these levels from Universe.t to something else. All that is fine (modulo the fact that these levels should not be calles sorts in the user doc, since sorts are another thing).

ppedrot:
This PR is basically an isomorphism so it shouldn't change the behaviour. If you want to preserve the previous user-facing name, you can push a patch atop of this PR, since I am not sure what you want in the end.

coq-builtin.elpi
```diff
@@ -792,7 +792,7 @@ external pred coq.env.primitive-projections i:inductive,
 
 % -- Universes --------------------------------------------------------
 
-% Univ.Universe.t
+% Sorts.t
```
gares, line 795:
```suggestion
% universe level
```

src/coq_elpi_HOAS.ml
```diff
@@ -152,16 +152,16 @@ let in_coq_fresh_annot_id ~depth ~coq_ctx dbl t =
 let univin, isuniv, univout, univ_to_be_patched =
   let { CD.cin; isc; cout }, univ = CD.declare {
     CD.name = "univ";
-    doc = "Univ.Universe.t";
+    doc = "Sorts.t";
```
gares, line 155:
```suggestion
    doc = "universe level";
```

SkySkimmer:
Please merge now

Issue #214, opened 2021-02-21
Redoing `Require Import elpi` breaks the parser


pi8027:
The following code fails
```
From elpi Require Import elpi.
Back.
From elpi Require Import elpi.

Elpi Db test.db lp:{{
test {{ lp:T }}.
}}.
```
with the following message:
```
$ coqtop < test.v
Welcome to Coq 8.13.0 (January 2021)

Coq < [Loading ML file ring_plugin.cmxs ... done]
[Loading ML file zify_plugin.cmxs ... done]
[Loading ML file micromega_plugin.cmxs ... done]
[Loading ML file omega_plugin.cmxs ... done]
[Loading ML file int63_syntax_plugin.cmxs ... done]
[Loading ML file float_syntax_plugin.cmxs ... done]
[Loading ML file elpi_plugin.cmxs ... done]

Coq <
Coq <
Coq < Coq < Coq < Coq < Toplevel input, characters 4-5:
Error: The reference T was not found in the current environment.
```
It works fine if I remove `Back.`

gares:
I know, and I'm afraid it is unfixable without elpi/elpi#13 see also #112.

PR #350, opened 2022-03-24
[elpi] use the new parser / elpi 1.15

gares:


pi8027:
Does this fix #214?

gares:
It should!

PR #353, opened 2022-03-25
Adapt w.r.t. coq/coq#15856.

ppedrot:


src/coq_elpi_builtins.ml
```diff
@@ -158,14 +160,23 @@ let add_universe_constraint state c =
...
 let mk_algebraic_super x = Sorts.super x
-let mk_algebraic_max x y = Sorts.sort_of_univ (Univ.Universe.sup (Sorts.univ_of_sort x) (Sorts.univ_of_sort y))
+let mk_algebraic_max x y =
+  let open Sorts in
+  match x, y with
+  | (SProp, SProp) | (Prop, Prop) | (Set, Set) -> x
+  | (SProp, (Prop | Set | Type _)) | ((Prop | Set | Type _), SProp) -> raise Pred.No_clause
+  | (Prop, (Set | Type _ as u)) | ((Set | Type _ as u), Prop) -> u
+  | (Set, Type u) | (Type u, Set) -> Sorts.sort_of_univ (Univ.Universe.sup u Univ.Universe.type0)
+  | (Type u, Type v) -> Sorts.sort_of_univ (Univ.Universe.sup u v)
```
gares, line 170: @ppedrot I'm not so fond of having CIC typing rules in Elpi. These should not be here.

gares (changes requested):

ppedrot:
@gares Are you fine if I replace `coq.univ.max` and `coq.univ.algebraic-max` with a function that computes the sort of a product?

src/coq_elpi_builtins.ml
ppedrot: The max of two levels doesn't exist either. It would be a `Universe.t`, i.e. some free algebraic expression. Elpi is confusing the three notions of level (a unique name), algebraic (list of levels + integer increments) and sorts (Prop, SProp, and Type@@{alg}). Either you expose the true complexity of the various objects to the user, or you hide it under a palatable interface based on sorts (the only real thing the user has access to in Coq) and you don't have max anymore but more refined notions. (By the way, sup is a lie as well, e.g. Prop + 1 is not Prop + 1 but Set + 1. This function computes the type of a sort.)

gares: I think I need to chat with you next week. I agree with what you say, but I wanted to give anothe illusion...

ppedrot: No problem. For now I don't have many constraints except Tuesday afternoon and the weekly call.

ppedrot:
@gares done. The test suite seems to pass.

elpi/elpi_elaborator.elpi
```diff
@@ -370,7 +370,7 @@ pred pts i:universe, i:universe, o:universe.
 
 pts  prop     prop     prop.
 pts (typ U)   prop     prop :- if (var U) (coq.univ.new [] U) true.
-pts (typ T1) (typ T2) (typ M) :- coq.univ.max T1 T2 M.
+pts (typ T1) (typ T2) (typ M) :- coq.univ.pts-triple T1 T2 M.
```
gares, line 373: @ppedrot I'm so predictable... I did not even recall I had a piece of code called `pts`.

gares:
Anyway, the new API should be lifted to `sort` rather than `univ`, but I won't delay this further.

SkySkimmer:
Please merge now

PR #356, opened 2022-03-29
Do not depend on Pretyping.known_glob_level.

ppedrot:
This function was used internally to Coq for constraint computations but was making other assumptions unrelated to the full generality of elpi.

The alternative is to export the `sort` function that is properly defined in Pretyping after #15863 but the advantage of this PR is that it is backwards compatible.

### DIFF
--- a/src/coq_elpi_glob_quotation.ml
+++ b/src/coq_elpi_glob_quotation.ml
@@ -87,6 +87,28 @@ let type_gen = ref 0
 
 let is_hole x = match DAst.get x with GHole _ -> true | _ -> false
 
+let universe_level_name evd ({CAst.v=id} as lid) =
+  try Evd.universe_of_name evd id
+  with Not_found ->
+    CErrors.user_err ?loc:lid.CAst.loc
+      (Pp.(str "Undeclared universe: " ++ Id.print id ++ str "."))
+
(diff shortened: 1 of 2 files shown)


src/coq_elpi_glob_quotation.ml
```diff
@@ -87,6 +87,23 @@ let type_gen = ref 0
...
+
+let sort env sigma l = match l with
+| [] -> assert false
+| [GSProp, 0] -> Sorts.sprop
+| [GProp, 0] -> Sorts.prop
+| [GUniv u, 0] -> Sorts.sort_of_univ (Univ.Universe.make u)
+| [GLocalUniv l, 0] ->
+  let u = universe_level_name sigma l in
+  Sorts.sort_of_univ (Univ.Universe.make u)
+| _ ->
```
gares, line 104: Can you make this one more precise? It may match also because a new `GWhatEver` is introduced, while  here you really mean `_ :: _` I believe.

gares (approved):

src/coq_elpi_glob_quotation.ml
ppedrot: No, the increment is important also. It could be a 1 (that's the only case which is really handled currently in the Coq codebase.)

PR #362, opened 2022-04-15
Mention Coqtail in README

whonore:
As of whonore/Coqtail#278, Coqtail provides syntax highlighting for embedded Elpi code and the related vernacular commands (e.g., `Elpi Command`), and correctly parses `lp:{{ ... }}` blocks. I also left the old syntax highlighting suggestion as an alternative, although it seems pretty incomplete.


gares:
Thanks! I think you could just remove the old vim related text, since I don't test it anymore and coqtail seems to cover syntax highlighting as well.

gares (approved):

gares:
Thanks!

PR #367, opened 2022-06-10
command Elpi Trace Browser

gares:
This PR adds a new command `Elpi Trace Browser`.

@jwintz If you compile this one (just make) then `ALT -> ` (coq: interpret to point) in file `examples/tutorial_elpi_lang.v` just after the new command, then start the trace watcher, then interpret the next Coq line, you should get a trace.

TODO:
- [ ] doc


gares:
I'm merging it so that we have it in the code. I leave the doc/announcement out waiting for the windows support to work (which may require changing this little patch, especially the `/tmp/traced.json.tmp` hardcoded file name

Issue #387, opened 2022-09-01
coq.env.add-indt fails to add universe polymorphic inductives when not in a section


ecranceMERCE:
Declaring a new universe polymorphic inductive type with `coq.env.add-indt` fails when not in a section.
```coq
From elpi Require Import elpi.
Set Universe Polymorphism.

Elpi Tactic test.

(* Section Test. *)

Elpi Query lp:{{
  coq.univ.new U,
  coq.univ.variable U L,
  @udecl! [L] tt [] tt =>
    coq.env.add-indt (inductive "MyFalse" tt (arity (sort (typ U))) (i\ [])) I.
}}.
```
```
No open section.
```
Uncommenting the line before the Elpi query makes the error vanish.

ecranceMERCE:
Issue https://github.com/LPCIC/coq-elpi/issues/374 might be related as the error message is the same.

PR #388, opened 2022-09-06
Fix #386

### LINKED ISSUE #386: Failure of coq.env.add-indt in the presence of non uniform parameters
Here is an inductive type:
```coq
Inductive t (n : nat) : Type :=
  | K : t O -> t n.
```
Here is its parametricity translation built in Elpi:
```elpi
inductive tR tt 
 (parameter n explicit (pglobal (indt «nat») «») c0 \
   parameter n' explicit (pglobal (indt «nat») «») c1 \
	parameter nR explicit (app [pglobal (indt «natR») «», c0, c1]) c2 \
     arity
      (prod `i` (app [pglobal (indt «t») «Test.78», c0]) c3 \
        prod `i'` (app [pglobal (indt «t») «Test.78», c1]) 
         c4 \ sort (typ «Test.78»))) c0 \
 [constructor KR 
   (parameter n explicit (pglobal (indt «nat») «») c1 \
     parameter n' explicit (pglobal (indt «nat») «») c2 \
      parameter nR explicit (app [pglobal (indt «natR») «», c1, c2]) c3 \
       arity
        (prod `_` 
          (app
            [pglobal (indt «t») «Test.78», 
             pglobal (indc «O») «»]) c4 \
          prod `_'` 
           (app
             [pglobal (indt «t») «Test.78», 
              pglobal (indc «O») «»]) c5 \
           prod `_R` 
            (app
              [c0, pglobal (indc «O») «», pglobal (indc «O») «», 
               pglobal (indc «OR») «», c4, c5]) c6 \
            app
             [c0, c1, c2, c3, 
              app [pglobal (indc «K») «Test.78», c1, c4], 
              app [pglobal (indc «K») «Test.78», c2, c5]]))]
```
Trying to add it to Coq through `coq.env.add-indt` fails with the following error:
```
3rd non uniform parameter, named nR, not found in constructor KR
```
Yet, I declared the corresponding inductive in Coq:
```coq
Inductive tR@{u} (n n' : nat) (nR : natR n n') : t@{u} n -> t@{u} n' -> Type :=
  | KR : forall (i i' : t O), tR O O OR i i' -> tR n n' nR (K n i) (K n' i').
```
and querying its declaration with `coq.env.indt-decl` gives *exactly* the same result that the one I wanted to add to Coq through Elpi. There is one universe per declaration which I all renamed to `u` to do a relevant `diff` on both versions, and only the naming changes as visible here:
```
6,7c6,7
<       (prod `i` (app [pglobal (indt «t») «u», c0]) c3 \
<         prod `i'` (app [pglobal (indt «t») «u», c1])
---
>       (prod `_` (app [pglobal (indt «t») «u», c0]) c3 \
>         prod `_` (app [pglobal (indt «t») «u», c1])
14c14
<         (prod `_`
---
>         (prod `i`
18c18
<           prod `_'`
---
>           prod `i'`
22c22
<            prod `_R`
---
>            prod `_`
```
and naming does not matter as everything is represented with Elpi binders.

ecranceMERCE:


gares:
From the log:
```
2022-09-06T16:46:41.1234079Z - File "./tests/test_invert.v", line 19, characters 0-25:
2022-09-06T16:46:41.1234435Z - Error:
2022-09-06T16:46:41.1234834Z - 1st non uniform parameter, named idx0, not found in constructor nilR_inv
2022-09-06T16:46:41.1235163Z - 
```
So it is not that simple. Maybe `lift 1` is simplistic, maybe it is a "lift from n_uniform_params"... but I had no time to dig into this

ecranceMERCE:
I took some time to investigate. I think I found the bug.
*Short: a lift is needed, but only for indices mentioning uniform parameters.*

Consider the following inductive type:
```coq
Inductive t (A : Type) (l : list A) (e : l = l) : Type :=
  | K : t A (@nil A) (@eq_refl A l) -> t A l e.
```
Here is its Elpi representation (simplified):
```elpi
parameter {{ Type }} A\
  inductive
    (parameter {{ list A }} l\ parameter {{ l = l }} e\ arity {{ Type }})
    t\ [
      constructor K
        (parameter {{ list A }} l\ parameter {{ l = l }} e\
          arity {{ t A (@nil A) (@eq_refl A l) -> t A l e }}
    ]
```
If we give this `indt-decl` value to `coq.env.add-indt`, before trying to add the inductive to Coq, it will call [`check_consistency_and_drop_nuparams`](https://github.com/LPCIC/coq-elpi/blob/e4bf0b0b3e092b16c89b4e49d64523902fe24d13/src/coq_elpi_HOAS.ml#L2684) to check that the non-uniform parameters found in the arity of the inductive and in the arity of each constructor are equal, but the check is performed *up to* consistency of the de Bruijn indices, because the arities of the constructors have an additional binder for the inductive.

IIUC, the previous solution to equate the indices was to shift by 1 all the indices in the arity of the inductive. But let's see the declaration with de Bruijn levels:
```elpi
parameter {{ Type }} λ.
  inductive
    (parameter {{ list #1 }} λ. parameter {{ #1 = #1 }} λ. arity {{ Type }})
    λ. [
      constructor K
        (parameter {{ list #2 }} λ. parameter {{ #1 = #1 }} λ.
          arity {{ ... }}
    ]
```
The two lists to compare are the following:
```elpi
[ {{ list #1 }}, {{ #1 = #1 }} ]
[ {{ list #2 }}, {{ #1 = #1 }} ]
```
As you can see, lifting all the indices in the first list makes the equality fail. IMO what we should do is lift *only* the indices that mention a value beyond the binder for the inductive, because they are the only ones losing their meaning if we add an additional binder under them. The ones mentioning something "closer" are unaffected, so already equal in both compared arities.

If you had a pointer to a helper function able to do that selective, it would be great. Otherwise I can try to write it.

gares:
maybe https://coq.github.io/doc/v8.10/api/coq/Vars/index.html#val-liftn , otherwise you will have craft a subst, same file.

gares:
sorry for brevity, I'm busy

ecranceMERCE:
`test_arg_HOAS.v` still fails on my machine at the moment we try to declare a new polymorphic inductive, probably because of https://github.com/LPCIC/coq-elpi/issues/387, but it is unrelated to the present issue. Both `test_invert.v` and my code seem to work with the selective lift update.

gares:
this looks good now, maybe I'll test it a bit more and merge.

THANKS!

PR #395, opened 2022-10-12
API: redflags

gares:
I'm almost satisfied with the API, although it is a bit verbose.
@msoegtropIMC this PR should be of some use.


MSoegtropIMC:
@gares : I just wanted to say that I am making slowish (learning Elpi) but steady progress based on this PR. I am not there yet, but I don't see blockers right now. The functions I need seem to work even with mutual fixpoints (after I found `coq.env.opaque?`). I can't say anything on performance as yet.

MSoegtropIMC:
@gares : digging a bit deeper, there is one issue: one can't list modules which contain module functors. E.g.:
```
Require Import compcert.lib.Maps.

Elpi Command modules.

Elpi Query lp:{{ coq.locate-module "compcert.lib.Maps" MP, coq.env.module MP L}}.
```

gares:
The problem is that I cannot give you a handle to objects behind a functor, since these objects don't exist (without instantiating the functor they don't have a type, for example).

IMO the `coq.env.module` API should be refined to return a list of stuff like:
```prolog
kind module-item type.

gref gref -> module-item.
module  modpath -> list module-item -> module-item.
modtype modtypath -> list module-item -> module-item.
module-functor  modpath -> module-item.
modtype-functor modtypath -> module-item.
```
where a `module` and `modtype` nodes are recursive as before, but the others are just "opaque".

MSoegtropIMC:
I guess this would do. Alternatively it would also be OK to return items in fully instantiated modules and ignore functors - for cbv purposes I am not really interested in non instantiated functors, since one can't use non instantiated symbols in cbv delta lists.

gares:
Not 100% happy, but here it is. I think I'll add a node `nIY` for the stuff which I can't represent correctly.

MSoegtropIMC:
@gares : I think the module API does what I want now, but I have difficulties filtering for `gref` entries - likely an Elpi basics issue. Why does the following code return an empty list for LG - I think it should be L filtered for gref entries:
```
Require Import compcert.lib.Maps.

Elpi Command modules.

Elpi Program simplifier lp:{{
  pred module-item-filter-gref i:module-item, o:gref.
  module-item-filter-gref (gref X) X.
}}.
Elpi Typecheck.

Elpi Query lp:{{ coq.locate-module "compcert.lib.Maps" MP, coq.env.module MP L, std.map-filter L module-item-filter-gref LG }}.
(* MP and L is what I would expect, but LG is empty and I don't understand why. *)
```

gares:
The problem is a bit weird, the warning at Elpi Typecheck is misleading.
```
Elpi Command modules.

Elpi Accumulate lp:{{
  pred module-item-filter-gref i:module-item, o:gref.
  module-item-filter-gref (gref X) X.
}}.
Elpi Typecheck.
```
`Elpi Program` is really low level, it does not load stuff. I don't know exactly why the warning is so vague.

I'll try to dig, but the code above works

MSoegtropIMC:
Interstingly this shows the module types but not the grefs:
```
Require Import compcert.lib.Maps.

Elpi Command modules.

Elpi Program simplifier lp:{{
  pred module-item-filter-gref i:module-item, o:module-item.
  module-item-filter-gref (gref X) (gref X) :- coq.say "OPT1" X, !.
  module-item-filter-gref (module-type X) (module-type X) :- coq.say "OPT2" X, !.
}}.
Elpi Typecheck.

Elpi Query lp:{{ coq.locate-module "compcert.lib.Maps" MP, coq.env.module MP L, std.map-filter L module-item-filter-gref LG }}.
```
=>
```
  LG = [module-type «compcert.lib.Maps.TREE», 
 module-type «compcert.lib.Maps.MAP», 
 module-type «compcert.lib.Maps.INDEXED_TYPE», 
 module-type «compcert.lib.Maps.EQUALITY_TYPE»]
```
I thought it might be some sort of name clash on gref and tried renaming gref to module-gref in this context (https://github.com/MSoegtropIMC/coq-elpi/commit/6374269c448c1bfff1ab3583048cc660726d04a8) but this did not help.

MSoegtropIMC:
P.S.: you talk about type checker warnings, but I don't see any in VsCoq. I usually see Elpi type checker errors. Should I use CoqIDE?

gares:
There is the same issue, you start a command and then a program.

gares:
> P.S.: you talk about type checker warnings, but I don't see any in VsCoq. I usually see Elpi type checker errors. Should I use CoqIDE?

They are attached to `Typecheck` "last character". Not very visible.

MSoegtropIMC:
> They are attached to Typecheck "last character". Not very visible.

Ahh, hovering over the tilde below the dot!

MSoegtropIMC:
> I'll try to dig, but the code above works.

It doesn't give type checker errors, but it still results in an empty LG.

gares:
This one works here:
```coq
Elpi Command modules.

Elpi Accumulate lp:{{
  pred module-item-filter-gref i:module-item, o:module-item.
  module-item-filter-gref (gref X) (gref X) :- coq.say "OPT1" X, !.
  module-item-filter-gref (module-type X) (module-type X) :- coq.say "OPT2" X, !.
}}.
Elpi Typecheck.

Elpi Query lp:{{ coq.locate-module "Ranalysis5" MP, coq.env.module MP L, std.map-filter L module-item-filter-gref LG }}.
```

gares:
I don't know what happens when you mix `Elpi Program` and `Elpi Command` + `Elpi Query`. I seems something very odd happens. Note that in my code there is no `Elpi Program`.

MSoegtropIMC:
@gares: sorry I think the issue is that the CompCert module I tested simply doesn't have grefs in the module itself, only in submodules. I see a lots and lots of grefs in the dump of L, but they seem to be all in the data for sub-modules.

I can't say I understand the difference between `Command` and `Program`. I guess I have to go over the tutorials again with this in mind.

MSoegtropIMC:
I can confirm that it works fine (so far) - sorry for the fuzz. I started out with module paths where I know that they contain grefs, but then in then when I hit the issue with module functors, I tested with variations of the module name and forgot to undo it.

gares:
`Program` does not accumulate coq specific code. I use it in the tutorial of Elpi itself, but as soon as you do Coq stuff you should use Command or Tactic.

MSoegtropIMC:
@gares : just wanted to let you know that I am still in the progress of putting everything together. So far all is fine. I expect that I am finished (in the sense that the complete tactic does what I want) by tomorrow.

MSoegtropIMC (approved):
I tested this PR in my development and it works fine. In the source review the only point I found is the documentation of the delta flag.

MSoegtropIMC:
@gares : one very minor issue I came across today: when I `coq.say` a `coq.redflags`(not `coq.redflag`) it prints as `TODO`.

The API is quite nice for my application. I store a coq.redflags item with the basic flags and the module symbols in a database once and then retrieve it and add the local symbols with `coq.redflags.add` in the reduction tactic. This way I get down to 15..20ms for the complete reduction tactic call (compared to 70s with simpl, so about 4000x faster).

gares:
Great. About the printing, I think I need to patch Coq in order to be able to print that datum. Does Ltac2 print it better? Maybe I've missed the right API.

MSoegtropIMC:
Ltac2 doesn't have a generic printer - it can print Coq terms but redflags aren't Coq terms. But then the redflags structure is not opaque in Ltac2 as it is in Elpi, so one could write a nicer printing function if one would feel the need for it in Ltac2. As far as I can see this is not so in Elpi. But this is not really much of an issue.

The reason I came across this is that I keep a database of the static part of a reduction (the basic flags + the module derived delta symbols). Then I dynamically add the local delta symbols to this reduction. This is quite convenient. The only issue is that I can't print it any more - I have to derive the content (delta symbol list) from the history of what I added.

PR #398, opened 2022-10-29
Adapt w.r.t. coq/coq#16754.

ppedrot:
Should be backwards compatible.


gares (approved):

ppedrot:
@gares ping.

gares:
Doesn't it also normalize the evar context?

ppedrot:
The new code is 100% compatible with the old one. (It's even definitionally equal to that, if you were to define this in OCaml.)

Issue #401, opened 2022-11-18
Universe polymorphic record declaration fails in presence of a coercion


ecranceMERCE:
Trying to add a `Box` record through Coq-Elpi fails if the field is a coercion. This issue is in the context of PR #400.

```coq
From elpi Require Import elpi.

Set Universe Polymorphism.

Elpi Command x.
Elpi Accumulate lp:{{
  gen :-
    coq.univ.new U,
    coq.univ.variable U I,
    RDecl =
      record "Box" (sort (typ {coq.univ.super U})) "MkBox" (
        field [coercion regular] "unbox" (sort (typ U)) (_\
      end-record)),
    coq.say "RDecl =" RDecl,
    @udecl! [I] tt [] tt => coq.env.add-indt RDecl _.
}}.
Elpi Typecheck.

Fail Elpi Query lp:{{ gen. }}.

Record Box@{i} : Type@{i+1} := MkBox { unbox :> Type@{i} }.
```

Changing `coercion regular` to `coercion off` works, but `coercion reversible` fails too.

cc @CohenCyril

CohenCyril:
Cc @gares

PR #404, opened 2022-11-21
:bug: Fix declaration of upoly record projections

### LINKED ISSUE #402: Record fields are not visible from outside the module if the record is universe polymorphic
I try to declare the following record with Coq-Elpi in a module `Test`:
```coq
Record Rec : Type := BuildRec { f : Type }.
```
```coq
Set Universe Polymorphism.

Elpi Command x.
Elpi Query lp:{{
  coq.env.begin-module "Test" none,
  Decl = record "Rec" {{ Type }} "BuildRec" (field [] "f" {{ Type }} (_\ end-record)),
  % @univpoly! =>
    coq.env.add-indt Decl _,
  coq.env.end-module _.
}}.

Print Module Test.
Check Test.f.
```
The `Print` command gives this:
```coq
Module Test := Struct
  Record Rec : Type := BuildRec { f : Type }.
  Definition f : forall _ : Rec, Type.
End
```
and the `Check` succeeds.

Now uncomment the option to make the record polymorphic and boom:
```coq
Module Test := Struct Record Rec@{u u0} : Type := BuildRec { f : Type }. End
```
The field `f` is not exported and the `Check` now fails.

cc @CohenCyril

ecranceMERCE:
Seems to fix #402

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -1863,16 +1863,17 @@ Supported attributes:
      if not (is_mutual_inductive_entry_ground me sigma) then
        err Pp.(str"coq.env.add-indt: the inductive type declaration must be ground. Did you forge to call coq.typecheck-indt-decl?");
      let primitive_expected = match record_info with Some(p, _) -> p | _ -> false in
-     let ubinders = 
+     let (uentry, uentry', ubinders) = 
        let open Entries in
        match me.mind_entry_universes with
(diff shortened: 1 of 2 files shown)


gares (approved):

ecranceMERCE:
Seems to fix #401 as well (the example now works).
```coq
About unbox.
```
```
unbox is a coercion
```

PR #406, opened 2022-11-24
api: coq.mod[ty]path->library

gares:
CC @msoegtropimc


MSoegtropIMC (approved):
LGTM and would solve the issues I mentioned in Zulip. Thanks a lot for the super fast support!

PR #417, opened 2022-12-31
cache_abbrev_for_tac: use GADT instead of Obj.magic and rectypes

SkySkimmer:


src/coq_elpi_builtins.ml
```diff
@@ -859,12 +859,13 @@ type tac_abbrev = {
   tac_fixed_args : Coq_elpi_arg_HOAS.Tac.glob list;
 }
 
+type ('a,'d) gbpmp = Gbpmp : ('d, _, 'b, Loc.t -> 'd) Pcoq.Rule.t * ('a -> 'b) -> ('a,'d) gbpmp
```
SkySkimmer, line 862: BTW what does gbpmp mean?

gares: Boh, probably god bless PMP

PR #424, opened 2023-01-16
arg_syntax: replace Pcoq.Entry.create with Pcoq.Entry.make

artagnon:
We are removing the deprecated Pcoq.Entry.create. See coq/coq#17065.


ppedrot:
@gares if it's normal that the CI fails, this should go in.

PR #427, opened 2023-01-22
Various small improvements to NES

swasey:


apps/NES/theories/NES.v
```diff
@@ -65,3 +65,41 @@ Elpi Accumulate lp:{{
 }}.
 Elpi Typecheck.
 Elpi Export NES.Open.
+
+(* List the contents a namespace *)
+Elpi Command NES.List.
```
swasey, line 70: We're not entirely satisfied with two commands. Bike-shedding welcome.

apps/NES/elpi/nes.elpi
```diff
@@ -152,6 +152,27 @@ current-path NS :- std.do! [
   std.rev Stack NS,
 ].
 
+% Find an existing namespace, or panic.
```
swasey, line 155: * We could use some examples. The use case here is `NES.Begin A. NES.Begin B. (* ... *) NES.End B. NES.Open B. (* rather than A.B *)`.
* The choice of `.Path` for "use the top-level `Path`" interacts poorly with the syntax of Elpi commands (e.g., one must write `NES.List ".Path".`).

apps/NES/Makefile.coq.local
```diff
@@ -0,0 +1,3 @@
+install-extra::
```
swasey, line 1: It might be nicer to move all of the predicates into `NES.Db`, rather than install `nes.elpi` because
* clients who want to use the predicates need two `Elpi Accumulate` commands rather than one, plus an `Extra Dependency`
* (minor) BedRock's version of `coqdep` does not seem to understand `From elpi.apps Extra Dependency "nes.elpi" as nes.`

gares:
Thanks for contributing back your code! I'm still travelling back from Boston, I'll look at the code ASAP.

gares:
Sorry, something came up and I could not look into this (and next days are already a mess for me).
CC @CohenCyril which knows NES very well, he might have good remarks esp. about the design of the new commands.

gares:
Thanks! The code looks great. What I can propose for "relative path" is to make `_.X` work as ident (exactly as `a.X` would, and as `_` works as well).
In this way you could write `NES.Open _.A` in place of `NES.Open ".A"`. What do you think? You can make the change yourself (and test it as part of this PR).
https://github.com/LPCIC/coq-elpi/blob/834323a62682db872eec95ced06d4b273cbe3400/src/coq_elpi_arg_syntax.mlg#L89-L90

gares:
or you can also do it another PR, no problem, but I guess you don't want to have `".X"` in your code base to begin with.

gares:
I'm planning an imminent release for 8.17, and I'd like to pull that in.
I think I can implement the `".X"` to `_.X` change myself, but would this change please you?

gares:
I mean, `_.P` would mean top level `P`, while otherwise `P` would mean relative to the current NS stack.
I'm also fine with `".P"`, but as you said it is a bit ugly, syntactically speaking.

swasey:
Sorry about the lag. I've been working on an internal deadline. I can spend time today (and tomorrow if necessary) on finishing touches.

* I liked and implemented the `_.P` change. FTR, I don't think we _have_ to change how argument `_` gets parsed.

* What do you think about eliminating `nes.elpi` and folding it into the database in module `NES`? I argued for that change in terms of ease of use for folks using NES (as a library) from Elpi code, but the change also avoids apparent coqdep and dune issues related to external dependencies.

* I think it makes sense to squash the various commits here prior to merging. Please let me know what you prefer.

gares:
I'm fine with you force pushing, on my side I just had this little fix you may want to integrate.
```diff
commit e42a375bce20ee82f481cb3896c972de7083e9bf (HEAD -> nes-improvements)
Author: Enrico Tassi <Enrico.Tassi@Inria.fr>
Date:   Thu Feb 9 13:34:36 2023 +0100

    fix test

diff --git a/apps/NES/_CoqProject.test b/apps/NES/_CoqProject.test
index a72d8e0..b5c9397 100644
--- a/apps/NES/_CoqProject.test
+++ b/apps/NES/_CoqProject.test
@@ -3,6 +3,7 @@
 -I ../../src
 -docroot elpi.apps
 
+-Q elpi     elpi.apps.NES
 -R theories elpi.apps
 -R tests    elpi.apps.NES.tests
 -R examples elpi.apps.NES.examples
diff --git a/apps/NES/tests/test_NES_lib.v b/apps/NES/tests/test_NES_lib.v
index 6138320..b3ada6a 100644
--- a/apps/NES/tests/test_NES_lib.v
+++ b/apps/NES/tests/test_NES_lib.v
@@ -1,4 +1,4 @@
-From elpi.apps Extra Dependency "nes.elpi" as nes.
+From elpi.apps.NES Extra Dependency "nes.elpi" as nes.
 From elpi.apps Require Import NES.
 
 Elpi Command Make.
```

About using a database, I'm not sure I understand your use case, especially given that the external dependency seems to work here (but I'm not using dune).

gares:
In any case, make a second database, eg `nes.db` (a proper data base) and `nes.source_code` or something like this (to make it clear it is not a "database").

I know a database is exactly like a file, but I'd like to distinguish this usage. Maybe it deserves a proper directive, and one day I'll add it, so I'd like to distinguish it. Internally a database is considered to be a non-static set of rules, hence it is "recompile" at every run of your program. Here you only need to name a piece of code (stored in the vo file) to avoid having to distribute a .elpi file, which is ok (even if distributing the file seems to work here).

swasey:
Thanks for spelling out the distinction between databases and files. I retract my proposal to drop the nes.elpi file. (I'd prefer to work around/fix the tool problems we've seen than "misclassify" the contents of nes.elpi.)

> About using a database, I'm not sure I understand your use case, especially given that the external dependency seems to work here (but I'm not using dune).

While I haven't narrowed the problems we saw down enough to file issues, here are our notes so far. (Just in case you are curious.)

1. coqdep seems to treat `External Dependency` unlike coqc and coqtop; specifically, it generates warnings about elpi files in opam/.../lib/coq/user-contrib
2. dune seeds its _build directory with .v files, and then runs coqdep and coqc using those copies; specifically, any .elpi file mentioned via `External Dependency` will not show up under _build

(1) matters to us because we stop our builds on coqdep warnings. (Easy to work around.)

(2) prevents us from using External Dependency targeting .elpi files in our own repos. (We can forego using .elpi files locally, as well as file a dune issue.)

swasey:
Aside: I noticed my original patch for `_` arguments didn't work under nix-shell and tweaked it.

Please feel free to merge or suggest other cleanup. It looks good to me.

swasey:
> I don't know dune super well, but the sources, including.elpi files, should be copied in _build.

Agreed. (We'll file an issue/PR with dune separately.)

> Still I don't know which -Q options are passed to coqc, see my patch. Maybe the _build/log contains the answer to my question

What question?

Note that I folded your -Q patch into the commit.

gares:
I meant that if you build with dune and it does not find the elpi file it may be that the directory which contains the file is not mapped to a logical path via -Q. I guess you can see how coqc is invoked by dune in the log file. Anyway, OT for this pr.

swasey:
> I noticed my original patch for _ arguments didn't work under nix-shell and tweaked it.

Looking again, I'm not yet convinced the `_` arguments tweak works with the right versions of Coq.

- I wrote and tested the original patch against 8.16.0
- nix-shell's older version of Coq (8.16+rc1) generated a warning
- I revised the patch to avoid the warning but did not test it

What version of Coq (or revision from Coq's repo) should I check against?

Meanwhile, I'll test things against 8.16.0.

swasey:
> What version of Coq (or revision from Coq's repo) should I check against?

Or, of course, I could hand off to you. You know more about the .mlg file than I do.

swasey:
> > I noticed my original patch for _ arguments didn't work under nix-shell and tweaked it.
> 
> Looking again, I'm not yet convinced the `_` arguments tweak works with the right versions of Coq.

Indeed the revised patch doesn't work against 8.16.0.

I will make it work again (at least for that Coq version) and add a test to catch regressions.

swasey:
> I will make it work again (at least for that Coq version) and add a test to catch regressions.

Done. The new test is test_NES_resolve.v.

gares:
OK, I'll take it from here tomorrow (hopefully)

PR #429, opened 2023-01-30
Adapt w.r.t. coq/coq#17170.

ppedrot:
This assumes that info_of_evar is only ever called on undefined evars, but I am not really confident this is the case. The test-suite goes through, though.


gares (approved):

ppedrot:
(This is backwards compatible.)

PR #433, opened 2023-02-09
Master on 8.17

gares:
TODO
- [ ] override elpi version in nix


gares:
@CohenCyril how do I pick a newer elpi in the config file? Can I override it from there?

PR #439, opened 2023-03-02
updating the CI

CohenCyril:
Restoring nix CI
(Discard result of CI jobs tagged `pull_request_target`: this is the old workflow running)
CC @gares


CohenCyril:
@gares there is an error but it's not a CI setup error, it looks like a real regression failure :grimacing:

PR #441, opened 2023-03-04
Clarify difference between coq.reduction.{vm,native}.norm

Blaisorblade:
I could be wrong but I guessed from the name and from the source code. Feel free to rephrase as appropriate. cc @jhaag


gares (approved):

gares (changes requested):
sorry, I'm ok with the changes, but this file is generated

Blaisorblade:
Oh right! I've fixed the original but I don't know the regeneration command — I've only found it for `src/coq_elpi_builtins_HOAS.ml` (now documented inline), and `git grep coq-builtin.elpi` finds nothing either. 
(And `make build` fails here since I have Coq 8.16).

gares (approved):

gares:
thanks

PR #443, opened 2023-03-20
Add TC for EqDec and Inj (from stdpp)

FissoreD:


gares:
we need to rebase on top of master, CI was fixed there.

gares:
@FissoreD I'm merging this into hook-tc. From now on please push to that branch (and remove this one as well).

PR #475, opened 2023-07-05
derive: set implicits globally

Blaisorblade:
Same bugfix as in #450, now extended to all remaining uses; this other use was introduced in #328 with the same bug.

~~Completely untested.~~ Now checks for this problem in all tests from #328.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -23,6 +23,10 @@
   the spilling of a predicate `{p}` needs to be accumulated after the
   type or mode of `p` is declared
 
+### APPS
+- `derive Inductive i {A}` now sets `A` implicit status globally
+- `lock Definition f {A}` now sets `A` implicit status globally
+
 ## [1.17.1] - 09/03/2023
 
 Requires Elpi 1.16.5 and Coq 8.17.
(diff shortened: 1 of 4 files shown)


gares:
Great!
Would you mind adding a line to the Changes file?

gares:
(I've fixed nix CI in master, if you could rebase that would give a green light as well)

Changelog.md
```diff
@@ -282,8 +286,8 @@ Requires Elpi 1.14.1 and Coq 8.15.
 Requires Elpi 1.13.6 and Coq 8.15.
 
 ### APPS
-- `derive Inductive i {A}` now correctly sets `A` implicit status
-- `lock Definition f {A}` now correctly sets `A` implicit status
+- `derive Inductive i {A}` now correctly sets `A` implicit status (XXX)
+- `lock Definition f {A}` now correctly sets `A` implicit status (XXX)
```
Blaisorblade, line 290: I can just drop this, but should we add a note here since we found later problems?

Blaisorblade:
I don't get the DOC failure — spurious? But the rest works!

gares (approved):

Changelog.md
Blaisorblade: In doubt, dropped again.

PR #477, opened 2023-07-07
[coq.typecheck] do not return a type since econstr API may have n…

gares:
…ormalized univs

- [ ] share code with coq.elaborate-skeleton


gares:
CC @ecranceMERCE

PR #503, opened 2023-09-19
Adapt to coq/coq#17836 (sort poly)

SkySkimmer:


src/coq_elpi_HOAS.ml
```diff
@@ -109,9 +109,12 @@ let add_universe_constraint state c =
   try add_constraints state (Set.singleton c)
   with
   | UGraph.UniverseInconsistency p ->
+      let sigma = (S.get (Option.get !pre_engine) state).sigma in
```
SkySkimmer, line 112: Is this the right way to get the evar_map? get_sigma is later in the file.

gares: yes

PR #508, opened 2023-09-28
Tactic argument for ltac value

### LINKED ISSUE #505: Get arguments for tactic of form `simple_intropattern_list(x)`
I need to write an Intro tactic that also allows for receiving normal intro patterns and also my own intro patterns. To be able to properly accept an intro pattern, I need to have a notation in the form of:
```coq
Tactic Notation "eiIntros" "(" simple_intropattern_list(x) ")" :=
  elpi eiIntros x.
```
amongst other notations.
It would thus be nice to be able to read in an intro pattern list, be able to manipulate it when necessary and pass it on to the intro tactic.

gares:
Fix #505

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## UNRELEASED
+
+### APIs
+
+- New `ltac1-tactic` opaque data type
+- New `tac` argument constructor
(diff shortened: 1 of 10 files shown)


gares:
That should be it @lukovdm

PR #522, opened 2023-10-17
TC solver 1

FissoreD:


gares:
@FissoreD can the new API have the same signature of the other one? Can you reconstruct the missing piece of information?

coq-builtin.elpi
```diff
@@ -1144,6 +1144,9 @@ external pred coq.TC.db-tc o:list gref.
 % [coq.TC.db-for GR Db] reads all instances of the given class GR
 external pred coq.TC.db-for i:gref, o:list tc-instance.
 
+% [coq.TC.db-for2 GR List Db] reads all instances of the given class GR
+external pred coq.TC.db-for2 i:gref, o:list gref.
```
gares, line 1148: I was talking about this API.
If you can return tc-instance instead of gref, we can just replace the old API above.

gares:
Maybe, but in some sense the value the use wrote is absent, hence the dummy zero.
It would be nice to get the right number indeed, as inferred by Coq.

gares:
If you manage to call the right API and complete the gref with the priority inferred by Coq, that would be super good and easy to document (the list is returned in order w.r.t. these numbers).

coq-builtin.elpi
FissoreD: Yep, I can do so

apps/tc/tests/importOrder/f3d.v
```diff
@@ -3,29 +3,30 @@ From elpi.apps.tc.tests.importOrder Require Import f2b.
...
 Module M4'.
   (* From elpi.apps.tc.tests.importOrder Require Import f2a. *)
   Elpi SameOrderImport.
 
   Global Instance f3a : A nat := {f x := x}.
-  Elpi AddInstances f3a.
+  
 
-  Section S1.
+  Section S1. Variable X : Type.
```
gares, line 16: do we still need the dummy (unused) variable here?

FissoreD: Nope, X can be removed

PR #525, opened 2023-10-21
Add "coq.env.query-extra-dep" external predicate

phikal:
As discussed on [Zulip](https://coq.zulipchat.com/#narrow/stream/253928-Elpi-users-.26-devs/topic/Resolving.20Extra.20Dependencies/near/396087392), this would allow for a predicate to resolve the real path behind a `Extra Dependency`.  As seen in the patch, this just amounts to wrapping `ComExtraDeps.query_extra_dep` in the right way.

### DIFF
--- a/coq-builtin.elpi
+++ b/coq-builtin.elpi
@@ -1666,6 +1666,10 @@ external pred coq.term->string i:term, o:string.
 % - @holes! (default: false, prints evars as _)
 external pred coq.term->pp i:term, o:coq.pp.
 
+% [coq.extra-dep Identifier File Name] Resolve the file name of an Extra
+% Dependency
+external pred coq.extra-dep i:id, o:option id.
+
 % -- Access to Elpi's data --------------------------------------------
 
 % clauses
--- a/coq-builtin.elpi
+++ b/coq-builtin.elpi
@@ -1666,8 +1666,10 @@ external pred coq.term->string i:term, o:string.
 % - @holes! (default: false, prints evars as _)
 external pred coq.term->pp i:term, o:coq.pp.
 
-% [coq.extra-dep Identifier File Name] Resolve the file name of an Extra
-% Dependency
+% -- Extra Dependencies -----------------------------------------------
+
+% [coq.extra-dep Identifier FileName] Resolve the file name of an extra
+% dependency. See also Coq's From xxx Extra Dependency yyy as zzz.
 external pred coq.extra-dep i:id, o:option id.
 
 % -- Access to Elpi's data --------------------------------------------
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -3888,8 +3888,8 @@ Supported attributes:
 
   MLCode(Pred("coq.extra-dep",
     In(id, "Identifier",
-    Out(option id, "File Name",
-    Easy "Resolve the file name of an Extra Dependency")),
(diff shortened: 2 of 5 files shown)


src/coq_elpi_builtins.ml
```diff
@@ -2325,6 +2325,15 @@ denote the same x as before.|};
             Some (c, np + na))))))),
   DocAbove);
 
+  MLCode(Pred("coq.env.query-extra-dep",
```
gares, line 2328: `coq.env` is for accessing the logical environment.
Please move this code just before 
```
-- Datatypes conversions --------------------------------------------
```
maybe in a dedicated section, say `-- Extra Dependency ------------------------`.

Also I'd call it `coq.extra-dep`

gares (changes requested):
Thanks for this PR.
There are a few things to improve before merging, see the comments.

You should also add a line to the ChangeLog.md file

phikal:
Enrico Tassi ***@***.***> writes:

> @gares requested changes on this pull request.
>
> Thanks for this PR.
> There are a few things to improve before merging, see the comments.
>
> You should also add a line to the ChangeLog.md file

Done.

>> @@ -0,0 +1,7 @@
> +From elpi Require Import elpi.
> +
> +From unreleased Extra Dependency "elpi_elaborator.elpi" as elab.
> +
> +Elpi Query lp:{{ coq.env.query-extra-dep "elab" File }}.
> +
> +Fail Elpi Query lp:{{ coq.env.query-extra-dep "foo" File }}.
>
> Please add the test file to _CoqProject.test.
>
> This file is a good test, but the API returns None instead of raising
> `No_clause`, so this will fail once it runs.

I've replaced it with an example that matches (some _).

>> @@ -2325,6 +2325,15 @@ denote the same x as before.|};
>              Some (c, np + na))))))),
>    DocAbove);
>  
> +  MLCode(Pred("coq.env.query-extra-dep",
>
> `coq.env` is for accessing the logical environment.
> Please move this code just before 
> ```
> -- Datatypes conversions --------------------------------------------
> ```
> maybe in a dedicated section, say `-- Extra Dependency ------------------------`.
>
> Also I'd call it `coq.extra-dep`

Fine with me, I've also changed this.

src/coq_elpi_builtins.ml
```diff
@@ -3884,6 +3884,15 @@ Supported attributes:
      state, !: s, [])),
   DocAbove);
 
+  MLCode(Pred("coq.extra-dep",
```
gares, line 3887: Please add an LPDoc line with something like
```
-- Extra Dependencies -------------
```

coq-builtin.elpi
```diff
@@ -901,6 +901,10 @@ external pred coq.env.projections i:inductive, o:list (option constant).
 external pred coq.env.primitive-projections i:inductive, 
                                             o:list (option (pair projection int)).
 
+% [coq.env.query-extra-dep Identifier File Name] Resolve the file name of an
```
gares, line 904: Please regenerate this file

gares (changes requested):
A few more nits

phikal:
Sorry, missed those.

gares (approved):

phikal:
Do you have a schedule when you are planning to publish a release with these changes?

gares:
Before the end of November

PR #536, opened 2023-11-08
fix installation of tc

Tragicus:
Modifies the `_CoqProject` files and `apps/tc/Makefile.coq.local` so that tc is installed in the `elpi/apps/tc` folder of the library.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -29,7 +29,7 @@
 -I apps/coercion/src
 
 # Type classes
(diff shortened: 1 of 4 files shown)


gares:
I suspect you need to change apps/tc/_CoqProject.tests

Tragicus:
Indeed, `make` now succeeds (at least locally).

gares:
thanks

PR #538, opened 2023-11-08
Class declare tc command

FissoreD:


apps/tc/elpi/create_tc_predicate.elpi
```diff
@@ -1,49 +1,76 @@
...
\ No newline at end of file
+add-class-str SearchMode ClassStr :-
+  coq.locate ClassStr ClassGR,
+  add-class-gr SearchMode ClassGR.
+
+% Following are predicates for TC.declare
+
+pred attr->deterministic o:search-mode. 
+attr->deterministic deterministic :- get-option "deterministic" tt, !.
+attr->deterministic classic.
```
gares, line 48: I think this should be called something like `search-mode`

PR #544, opened 2023-11-11
Missing API on universes

CohenCyril:
@gares @ecranceMERCE would it make sense to merge this upstream?

### DIFF
--- a/coq-builtin.elpi
+++ b/coq-builtin.elpi
@@ -957,6 +957,9 @@ external pred coq.univ.print .
 % [coq.univ.new U] A fresh universe.
 external pred coq.univ.new o:univ.
 
+% [coq.univ.super U U1] relates a univ U to its successor U1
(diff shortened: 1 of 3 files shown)


palmskog:
Given that [Trocq](https://github.com/coq-community/trocq) that needs this PR has joined Coq-community, could you comment on the feasibility of merging it in the near future @gares?

If it's not feasible, we may be faced with maintaining a fork of Coq-Elpi in Coq-community, which would be unfortunate for everyone involved. Other solution suggestions also welcome.

gares:
I'm all for helping out but this PR breaks ci, I did not even look into it to be honest, since I did not know there was any urgency

palmskog:
There is no huge urgency really, it's more about deciding whether we would need to fork Coq-Elpi in Coq-community just to be able to test and package Trocq. Maybe you can say something about when you might have time to take a look at the PR assuming it's not urgent, and then we can decide after that.

PR #555, opened 2023-11-28
namegen API

### LINKED ISSUE #554: Add binding for coqs namegen API
I need to generate a fresh name to introduce forall in a goal. However, I need to generate a fresh name for it. A binding for the api would allow me to get a fresh name to introduce with.

gares:
@lukovdm please check this is what you need. You can pass a default name or a type as the seed and the ID is fresh in the current context.

Fix #554


lukovdm:
Yes, that is exactly what I needed

PR #565, opened 2023-12-23
Next master

gares:


gares:
@CohenCyril do you know why the nix job picks up an old version of coq-elpi and not this branch?

.nix/config.nix
```diff
@@ -10,8 +10,6 @@
       hierarchy-builder.override.version = "coq-elpi-2";
       hierarchy-builder-shim.job = false;
 
-      coq-elpi.override.version = "master";
```
CohenCyril, line 13: @gares this was asking the action to test the master branch explicitly.

gares: Thanks

gares:
Thanks @CohenCyril I now see it passes MC and OOthm, so I'm OK releasing this branch.
But there is a problem with finmap, otherwise I'd love to also test MCA before releasing.

PR #567, opened 2023-12-29
[build] dune

gares:
blocked by [bug](https://github.com/ocaml/dune/issues/7091)
unblocks using ppx_optcomp and have a single branch for multiple coq versions


rlepigre:
I think that this can be unblocked with current dune by creating "dummy" theories for the various `elpi` folders. This requires changing the logical paths a bit, but that should probably not be a problem.

The only slight issue is that the top-level `elpi` folder cannot be put under a logical path with the `elpi.` prefix (like `elpi.elpi`) due to something I don't completely understand. But maybe that can be worked around.

Here is my experiment: https://github.com/rlepigre/coq-elpi/tree/dune (based on a rebased version of this branch). It seems to work as expected, but it is hard to tell given the amount of output that tests give. :sweat_smile:

PR #568, opened 2024-01-04
Update apps/tc/README.md

gares:


FissoreD:
> TC.Declare should declare "class methods" so that this works

In pure coq, we get the same error of unbound method name:
```coq
Record Eqb (T: Type) := {
    eqb : T -> T -> bool; 
    eqb_leibniz A B: eqb A B = true <-> A = B
}.

Existing Class Eqb.

#[refine] Instance eqBool : Eqb bool := {
  eqb x y := if x then y else negb y
}.
Proof. intros [] []; intuition. Qed.
```
the call to `Record.declare_existing_class` does not build the class mehods.

PR #571, opened 2024-01-08
[TC] generalize univ-instance

FissoreD:
- [] univ-instance is replaced with an elpi fresh var to ease unification
- [] allow multi trace with `Elpi Trace Browser`


apps/tc/theories/tc.v
```diff
@@ -105,6 +106,25 @@ Elpi Accumulate lp:{{
...
+Elpi Accumulate File tc_aux.
+Elpi Accumulate lp:{{
+  pred add-unfold i:gref.
+  add-unfold (const C) :-
+    if (unfold-constant C) true
+      (add-tc-db _ _ (unfold-constant C)).
+  add-unfold GR :-
+    coq.error "[TC]" GR "is not a constant".
+  main L :-
+    ErrMsg = "[TC] TC.Unfold accepts a list of string is accepted",
```
gares, line 122:
```suggestion
    ErrMsg = "[TC] TC.Unfold takes a list of strings",
```

PR #572, opened 2024-01-11
Expose `Replace grafting for coq.elpi.accumulate

FissoreD:
This PR work with [this](https://github.com/LPCIC/elpi/pull/217)

### DIFF
--- a/src/coq_elpi_builtins_synterp.ml
+++ b/src/coq_elpi_builtins_synterp.ml
@@ -53,7 +53,9 @@ let scope = let open Conv in let open API.AlgebraicData in declare {
 
 let grafting = let open Conv in let open API.AlgebraicData in declare {
   ty = TyName "grafting";
-  doc = "Specify if the clause has to be grafted before or after a named clause";
+  doc = {|Specify if the clause has to replace or to be grafted before or after a 
+named clause. 
+Note: replace is only possible for unnamed clauses |};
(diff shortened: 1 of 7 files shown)


src/coq_elpi_builtins_synterp.ml
```diff
@@ -53,7 +53,9 @@ let scope = let open Conv in let open API.AlgebraicData in declare {
 
 let grafting = let open Conv in let open API.AlgebraicData in declare {
   ty = TyName "grafting";
-  doc = "Specify if the clause has to be grafted before or after a named clause";
+  doc = {|Specify if the clause has to replace or to be grafted before or after a 
+named clause. 
+Note: replace is only possible for unnamed clauses |};
```
CohenCyril, line 58: -> "only for named clauses" or "impossible for unnamed closes" right?

FissoreD: I admit my sentence is not very clear, what I mean is the follwong:
we can replace `C1` with the new clause `C2` only if `C1` has a name . However, `C2` cannot have a name. 
For instance 
```prolog
coq.elpi.accumulate _ "my_db" (clause "HERE_THE_ERROR" (replace "graft_me2") (graft_me 3)).
```
raises an error since the new clause `graft_me 3` can't have a name. This comes from [here](https://github.com/LPCIC/elpi/blob/e1511235ae98f9b9a9d1c431f853277fcf406602/ELPI.md?plain=1#L246)

PR #577, opened 2024-01-24
CS app

Tragicus:
Canonical structure hook, with a copy of `evarconv.ml` so that we can make modifications locally and override Coq's current behavior.
N.B. This can evolve easily into an unification hook.
At this time, it is mostly untested.


gares (changes requested):
Thanks for your contribution!!!

There is file which should be removed (it is generated).
I'd also like `canonical-solution` to be shorted, maybe `cs` or `csolution`.

apps/cs/TODOREADME.md
```diff
@@ -0,0 +1,44 @@
...
+## The canonical-solution predicate
+
+The `canonical-solution` predicate lives in the database `cs.db`
+
+```elpi
+% predicate [canonical-solution Ctx Lhs Rhs] used to unify Lhs with Rhs, with
+% - [Ctx] is the context
+% - [Lhs] and [Rhs] are the terms to unify
+:index (0 6 6)
+pred canonical-solution i:goal-ctx, o:term, o:term.
```
gares, line 16: Fix and rename the file plz

apps/cs/tests/test_cs.v
```diff
@@ -0,0 +1,38 @@
...
+From Coq Require Import Bool.
+
+Elpi Override CS All.
+
+Structure S : Type :=
+  { sort :> Type }.
+
+Elpi Accumulate cs.db lp:{{
+
+canonical-solution _ {{ sort lp:Sol }} {{ nat }} :-
```
gares, line 11: Fix

Tragicus:
I do not understand what happens with the compilation. There is no "Replace" in the code and yet it magically appears in the .cmt file.

gares (approved):

PR #581, opened 2024-01-26
fix sigma pruning when calling ltac, reachability is not correct

gares:
to be checked: examples/example_abs_evars.v

### DIFF
--- a/src/coq_elpi_HOAS.ml
+++ b/src/coq_elpi_HOAS.ml
@@ -2434,10 +2434,10 @@ let solution2evd sigma0 { API.Data.constraints; assignments; state; pp_ctx } roo
     get_declared_goals (Evar.Set.diff all_goals solved_goals) constraints state assignments pp_ctx in
   debug Pp.(fun () -> str "Goals: " ++ prlist_with_sep spc Evar.print declared_goals);
   debug Pp.(fun () -> str "Shelved Goals: " ++ prlist_with_sep spc Evar.print shelved_goals);
-  Evd.fold_undefined (fun k _ sigma ->
+(* wrong:  Evd.fold_undefined (fun k _ sigma ->
     if Evar.Set.mem k all_goals || Evd.mem sigma0 k then sigma
     else Evd.remove sigma k
-    ) sigma sigma,
+    ) sigma*) sigma,
   declared_goals,
   shelved_goals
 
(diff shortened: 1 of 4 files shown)


gares:
@Tragicus this new test seems more correct. Can you try it in your example? (After all I can't completely remove the pruning as we were doing this afternoon).

src/coq_elpi_HOAS.ml
```diff
@@ -2429,13 +2432,14 @@ let reachable sigma roots acc =
...
+  let reachable_undefined_evars = reachable sigma roots Evar.Set.empty in
   let declared_goals, shelved_goals =
-    get_declared_goals (Evar.Set.diff all_goals solved_goals) constraints state assignments pp_ctx in
+    get_declared_goals (Evar.Set.diff reachable_undefined_evars solved_goals) constraints state assignments pp_ctx in
   debug Pp.(fun () -> str "Goals: " ++ prlist_with_sep spc Evar.print declared_goals);
   debug Pp.(fun () -> str "Shelved Goals: " ++ prlist_with_sep spc Evar.print shelved_goals);
   Evd.fold_undefined (fun k _ sigma ->
-    if Evar.Set.mem k all_goals || Evd.mem sigma0 k then sigma
+    if Evar.Set.mem k reachable_undefined_evars then sigma
     else Evd.remove sigma k
```
ppedrot, line 2443: Sorry for hijacking this PR, but you definitely shouldn't reimplement your own pruning in a plugin. As a matter of fact, I wonder why Evd.remove is not marked as deprecated in the Coq API as you can easily wreak havoc (and definitely will if people use additional Store primitives in an evarmap).

We should instead provide a pruning function directly with Evd and ask in the API when creating user-defined evar data to provide functions explaining how the new data creates pointers to other blocks. It is more than time to realize that an evarmap is a proper heap...

Tragicus:
> @Tragicus this new test seems more correct. Can you try it in your example? (After all I can't completely remove the pruning as we were doing this afternoon).

It works.

src/coq_elpi_HOAS.ml
gares: I'm all for a heap that comes with a garbage collector, but currently it does not.
Worse, there is code that insists all your allocations are pointed to by the root, that is, that you don't leave garbage.

I'm happy to use any API you provide.

After this is merged, we can see if it also fixes the other bug affecting the TC solver.

PR #582, opened 2024-01-26
add catches

Tragicus:


src/coq_elpi_vernacular.ml
```diff
@@ -639,7 +642,9 @@ let run_tactic_common loc ?(static_check=false) program ~main ?(atts=[]) () =
     | API.Execute.Success solution -> Coq_elpi_HOAS.tclSOLUTION2EVD sigma solution
     | API.Execute.NoMoreSteps -> CErrors.user_err Pp.(str "elpi run out of steps")
     | API.Execute.Failure -> elpi_fails program
-    | exception (Coq_elpi_utils.LtacFail (level, msg)) -> tclFAILn level msg)
+    | exception (Coq_elpi_utils.LtacFail (level, msg)) -> tclFAILn level msg
+    (*| exception (CErrors.UserError _ as e) -> Exninfo.iraise (Exninfo.capture e) *)
+    | exception e -> let e = Exninfo.capture e in (Feedback.msg_debug Pp.(str "elpi lets escape exception: " ++ CErrors.print (fst e)); Exninfo.iraise e))
```
gares, line 647:
```suggestion
    | exception e -> let e = Exninfo.capture e in (Feedback.msg_debug Pp.(str "elpi lets escape exception: " ++ CErrors.print (fst e)); Exninfo.iraise e))
```

gares:
we should check the TC, CS and coercion apps also catch these, so maybe we sound factor the code in Interp.run

src/coq_elpi_vernacular.ml
```diff
@@ -602,6 +602,9 @@ module Interp = struct
       end with e ->
         let e = Exninfo.capture e in
         Vernacstate.Synterp.unfreeze final_synterp_state;
+        (match fst e with
+         | CErrors.UserError _ -> ()
+         | _ -> Feedback.msg_debug Pp.(str "elpi lets escape exception: " ++ CErrors.print (fst e)));
```
gares, line 607:
```suggestion
```

(only for test)

PR #585, opened 2024-02-01
Supporting no purge

CohenCyril:


src/coq_elpi_HOAS.ml
```diff
@@ -1105,6 +1108,12 @@ let get_options ~depth hyps state =
       let _, rd, gl = reduction_flags.Elpi.API.Conversion.readback ~depth state t in
       assert (gl = []);
       Some rd in
+  let keeping_algebraic_universes s =
+    if s = Some "default" then None
+    else if s = Some "keep-alguniv" then Some true
+    else if s = Some "purge-alguniv" then Some false
+    else if s = None then None
```
gares, line 1115: maybe a bool would just do it then

src/coq_elpi_HOAS.ml
```diff
@@ -908,9 +910,10 @@ let purge_algebraic_univs_sort state s =
 
 let in_elpi_flex_sort t = E.mkApp sortc (E.mkApp typc t []) []
 
-let sort = { sort with API.Conversion.embed = (fun ~depth state s ->
+(* WIP: I do not know how to make this optional *)
```
gares, line 913: I guess the only way is to make this a ContextualConversion, in that case you receive a `{ option }` rather than just a state. I may try to fix that tomorrow if time allows for it.

src/coq_elpi_HOAS.ml
CohenCyril: What was the rationale  for choosing `bool option` over `bool` in almost all other attributes? E.g. `coq:keepunivs`, `coq:locality`, etc

gares:
the code is now ready, but I think something is messed up sine the test suite fails.
I need to check the entire patch to see what changed (in theory the change is backward compatible).

tests/test_API_env.v
```diff
@@ -191,7 +191,8 @@ Elpi Query lp:{{
            lp:t lp:y true }})
        , constructor "K2x" (parameter "y" _ {{nat}} y\ arity {{
            lp:t lp:y false }}) ]),
-  coq.typecheck-indt-decl D ok,
+           coq.say D,
```
gares, line 194:
```suggestion
```

gares:
now it passes the test suite, but I'd like some tests in the test suite.

gares:
can you push some?

gares:
I did add a test, renamed the new API, and added a changelog entry. If CI is green I'll merge

src/coq_elpi_HOAS.ml
gares: I did update it. My point was to have an option carrying a bool, not a string.

gares:
I think this was ok, but got stuck because CI was broken.
It now requires a rebase.

PR #591, opened 2024-02-08
Cswb

Tragicus:


apps/cs/src/evarconv_hacked.ml
```diff
@@ -591,6 +594,38 @@ let infer_conv_noticing_evars ~pb ~ts env sigma t1 t2 =
     if !has_evar then None
     else Some (UnifFailure (sigma, UnifUnivInconsistency e))
 
+let pr_econstr = ref (fun _ _ _ -> Pp.str "unable to print econstr")
+
+(* TODO: move to a proper place *)
```
gares, line 599: Look at the `CList` module (coq has an extended lib)

src/coq_elpi_builtins.ml
```diff
@@ -3382,6 +3382,21 @@ Supported attributes:
        !: t)),
   DocAbove);
 
+   MLCode(Pred("coq.reduction.lazy.whd_betaiota_deltazeta_for_iota_state",
+    CIn(term,"T",
+    COut(term,"Tred",
+    Read(proof_context, {|Puts T in weak head normal form.
```
gares, line 3388: I have no clue what it does, but it does not do "just whd" ;-)

Tragicus: Indeed, thank you. Is there a difference between "weak head normal form" and "head normal form"? If so, the documentation of `whd_betaiota_deltazeta_for_iota_state` in `reductionops.ml` states "head normal form", while the function name starts with "whd", is that a typo?

Tragicus: Also, would you like for me to PR this commit (after fixing the doc) to coq-elpi now?

PR #593, opened 2024-02-12
Update cs hook

Tragicus:
Updates the cs hook to send the right problem to elpi.

### DIFF
--- a/apps/cs/theories/cs.v
+++ b/apps/cs/theories/cs.v
@@ -3,18 +3,20 @@ From elpi Require Import elpi.
 
 Elpi Db cs.db lp:{{
 
-  % predicate [canonical-solution Ctx Lhs Rhs] used to unify Lhs with Rhs, with
+% predicate [cs Ctx Proj Rhs Sol] used to find Sol such that Proj Sol = Rhs, where
 % - [Ctx] is the context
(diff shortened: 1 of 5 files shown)


apps/cs/theories/cs.v
```diff
@@ -3,25 +3,26 @@ From elpi Require Import elpi.
...
-Elpi Accumulate lp:{{
+Elpi Accumulate Db cs.db.
+Elpi Accumulate canonical_solution lp:{{
 
-solve (goal Ctx _ {{ @eq lp:T lp:Lhs lp:Rhs }} _P []) _ :-
-  cs Ctx Lhs Rhs,
-  % std.assert! (P = {{ eq_refl lp:Lhs }}) "cs: wrong solution".
+solve (goal Ctx _ _Ty Sol [trm Proj, trm Rhs]) _ :-
+  cs Ctx Proj Rhs Sol,
+  % std.assert! (P = {{ eq_refl lp:Lhs }}) "cs: wrong solution".
```
gares, line 23: Is `Ty` the type of `Sol`? If so, it would be nice to document it here.

Tragicus: Should I not even remove it, since we are never going to use it ?

PR #597, opened 2024-02-15
Add synterp phase support for derive plugins.

rlepigre:
This was developed and tested against our own derive commands (included in [this branch](https://github.com/rlepigre/coq-elpi/tree/br/derive-synterp-dirty)). It would probably make sense to test this in the repo somehow though, maybe by adding a new deriver that creates some kind of module?

CC @Janno.

**Edit:**

We had to require a `synterp` phase for all derivers since we need to know about all of them upfront (the `dep1` arguments are not enough to compute the full list, since independent derivers have no `dep1`). This does not seem too bad though.

We tweaked the module wrapping behaviour of `derive` in the following way:
- Using `derive Inductive ty := ...` does not automatically create a module `ty` anymore.
  - The original behaviour can be recovered with `#[module] derive Inductive ty := ...`.
  - A module name can be specified with `#[module="Ty"] derive Inductive ty := ...`.
- A custom prefix can be configured with the `prefix` (string) attribute.
- When a wrapper module is generated, the `no_alias` (boolean) attribute can be used to avoid generating alias notations.

### DIFF
--- a/apps/derive/elpi/derive.elpi
+++ b/apps/derive/elpi/derive.elpi
@@ -89,53 +89,84 @@ pred indt-or-const i:gref.
 indt-or-const (indt _).
 indt-or-const (const _).
 
-pred main i:gref, i:bool, o:list prop.
-main T P CL :- get-option "recursive" tt, !,
-  coq.env.dependencies T _ AllDeps,
+pred main i:gref, o:list prop.
+main GR CL :- get-option "module" M, !,
+  if (M = "") (coq.gref->id GR Mod) (Mod = M),
+  if-verbose (coq.say "Starting module" Mod),
+  coq.env.begin-module Mod none,
+  main-derive GR tt CL,
+  coq.env.end-module _.
+main GR CL :-
+  main-derive GR ff CL.
+
+pred main-derive i:gref, i:bool, o:list prop.
+main-derive GR InModule CL :- get-option "recursive" tt, !,
+  coq.env.dependencies GR _ AllDeps,
   coq.gref.set.elements AllDeps AllDepsL,
   std.filter AllDepsL indt-or-const Deps,
-  main.aux Deps [] CL1,
-  CL1 => main1 T P CL2,
+  main.aux InModule Deps [] CL1,
+  CL1 => main1 GR InModule CL2,
   std.append CL1 CL2 CL.
-pred main.aux i:list gref, i:list prop, o:list prop.
-main.aux [] X X.
-main.aux [T|TS] Acc CL :-
-  (pi X\get-option "only" X :- !, fail) => Acc => main T tt CL1,
-  main.aux TS {std.append CL1 Acc} CL.
+main-derive GR InModule CL :- main1 GR InModule CL.
 
-main T P CL :- main1 T P CL.
+pred main.aux i:bool, i:list gref, i:list prop, o:list prop.
+main.aux _ [] X X.
+main.aux InModule [GR|GRS] Acc CL :-
+  (pi X\get-option "only" X :- !, fail) => Acc => main-derive GR InModule CL1,
+  main.aux InModule GRS {std.append CL1 Acc} CL.
 
+pred validate-recursive i:prop, o:derive.
+validate-recursive (derivation _ _ tt _) _ :- get-option "recursive" tt,
+  coq.error "Synterp actions not supported in recursive derive.".
+validate-recursive (derivation _ _ _ R) R.
 pred main1 i:gref, i:bool, o:list prop.
-main1 T P CL :-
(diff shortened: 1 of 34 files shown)


apps/derive/elpi/derive.elpi
```diff
@@ -89,47 +89,76 @@ pred indt-or-const i:gref.
...
+     (if (InModule is ff) (Prefix is {coq.gref->id GR} ^ "_") (Prefix = "")),
+  std.findall (derivation GR Prefix _ _) L,
   if (L = [])
      (coq.error "no derivation found, did you Import derive.std?")
      true,
-  std.map L (x\r\ x = derivation _ _ r) DL,
-  validate-only T DL,
+  std.map L (x\r\
+    if (get-option "recursive" _)
+       (x = derivation _ _ S r,
```
gares, line 129: I'm busy this morning, I'll review later, but this popped up. You need to `sigma S\` for this to work.
I'd recommend using a named predicate to validate and map the derivation.

gares:
> * Using `derive Inductive ty := ...` does not automatically create a module `ty` anymore.
>   
>   * The original behaviour can be recovered with `#[module] derive Inductive ty := ...`.
>   * A module name can be specified with `#[module="Ty"] derive Inductive ty := ...`.
> 
> * A custom prefix can be configured with the `prefix` (string) attribute.

please put this text in the README.md (and fix the example there).

apps/derive/elpi/derive.elpi
rlepigre: Good thing you noticed! This should be fixed.

apps/derive/elpi/derive.elpi
```diff
@@ -124,14 +133,7 @@ main1 GR InModule CL :-
...
      true,
-  std.map L (x\r\
-    if (get-option "recursive" _)
-       (x = derivation _ _ S r,
-        if (S is tt)
-           (coq.error "Synterp actions not supported in recursive derive.")
-           true)
-       (x = derivation _ _ _ r)
-  ) DL,
+  std.map L (x\r\ validate-recursive x r) DL,
```
gares, line 136:
```suggestion
  std.map L validate-recursive DL,
```

apps/derive/elpi/derive.elpi
```diff
@@ -115,6 +115,15 @@ main.aux InModule [GR|GRS] Acc CL :-
...
 
+pred validate-recursive i:prop, o:derive.
+validate-recursive X R :-
+  if (get-option "recursive" _)
+     (X = derivation _ _ S R,
+      if (S is tt)
+         (coq.error "Synterp actions not supported in recursive derive.")
+         true)
+     (X = derivation _ _ _ R).
+
```
gares, line 126:
```suggestion
validate-recursive (derivation _ _ tt R) _ :- get-option "recursive" tt,
  coq.error "Synterp actions not supported in recursive derive.".
validate-recursive (derivation _ _ _ R) R.
```

apps/derive/elpi/derive_synterp.elpi
```diff
@@ -0,0 +1,80 @@
...
+tgt-is-not-in [_|L] D :- tgt-is-not-in L D.
+
+pred topo i:list derive, i:list prop, o:list derive.
+topo [] _ [] :- !.
+topo L Deps SL :-
+  std.partition L (not-a-src Deps) LNoDeps Other,
+  if (LNoDeps = []) (coq.error "derive: no topological order:" L Deps) true,
+  std.filter Deps (tgt-is-not-in LNoDeps) NewDeps,
+  topo Other NewDeps SOther,
+  std.append LNoDeps SOther SL.
```
gares, line 56: TODO (for me): there is a coq.elpi.toposort but for a full graph. Fix/adapt to this use case.

rlepigre:
Note: I fixed the generation of the alias for the type in the case where there is a wrapper module. It was named after the wrapper module name, instead of the type name.

I also took this opportunity to add a `no_alias` attribute that disables the generation of the aliases.

rlepigre:
> please put this text in the README.md (and fix the example there).

@gares I have given it a quick try, and it seems to me like the `README.md` file is not in a terribly good shape right now. I'd be surprised if the code it contains fully works on master (I found a few typos). I'm not sure how best to fix this.

gares:
Thanks. The code should be copy/pasted from readme.v which you edited to add the `#[module]` attribute and that edit prompted for an update. The README says it generates a module, that now is optional.

rlepigre:
I took care of adding a bit of documentation to `apps/derive/README.md`, and also squashed commits since they did not really make sense anymore on their own with all the bug fixes we pushed.

gares (approved):

PR #602, opened 2024-02-22
API to print a goal

gares:
still rough, but a starting point
CC @gstew5


gstew5:
Thanks for this, @gares!

PR #604, opened 2024-02-29
Structured synterp actions

Janno:
This implements are more structured log representation for synterp actions. In particular, it allows the creation of groups around multiple actions (and around other groups) that can be replayed all at once. See the NES changes for examples of how to use it.

This removes the `replay-all-missing-synterp-actions` predicate since it is not very compositional and the new group structure allows for more fine-grained control over replaying actions.

The code is still in rough shape in some places but it would be great to get feedback early.


gares:
I like the idea. I'll try to look at the code tomorrow

Janno:
FYI: @rlepigre (who wrote the code with me) suggested to return an opaque identifier from `begin-synterp-group` that would be given to `end-synterp-group` so that users do not have to keep strings in sync in order to ensure well-bracketedness. We'll probably implement that soon. It also removes the possibility of accidentally closing somebody else's synterp group that happens to have the same name.

rlepigre:
FYI: I implemented the unique identifier approach, and did some interface cleanup.

coq-builtin-synterp.elpi
```diff
@@ -363,12 +363,13 @@ type export-module modpath -> synterp-action.
...
-external pred coq.begin-synterp-group i:id.
+% [coq.begin-synterp-group ID Group] Create and open a new synterp action
+% group with the given name.
+external pred coq.begin-synterp-group i:id, o:group.
 
-% [coq.end-synterp-group ID] End the synterp action group ID. ID must be the
-% current group.
-external pred coq.end-synterp-group i:id.
+% [coq.end-synterp-group Group] End the synterp action group Group. Group
+% must refer to the most recently openned group.
```
gares, line 371:
```suggestion
% must refer to the most recently opened group.
```

src/coq_elpi_builtins_synterp.ml
```diff
@@ -809,31 +829,34 @@ module SynterpAction = struct
...
-        match ReadZipper.open_group rz name with
-        | Ok rz -> rz
-        | Error `Empty ->
-          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match a begin-synterp-group operation on group " ^ name ^ " but there are no groups or actions left in the current state.")))
-        | Error (`GroupMismatch name') ->
-          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match a begin-synterp-group operation on group " ^ name ^ " but the next group to match has name " ^ name' ^ ".")))
-        | Error `LeadingAction ->
-          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match a begin-synterp-group operation on group " ^ name ^ " but the next item to match is an action, not a group.")))
-      ), (), [])),
+      Out(Tree.group, "Group",
```
gares, line 832: You also want to declare a `MLData Tree.group` so that it shows up in the generated file.

README.md
```diff
@@ -290,11 +290,16 @@ The synterp-command can output data of that type, but also any other data it
...
 
 The second way to communicate data is implicit, but limited to synterp actions.
-During the interp phase commands can use the `coq.next-synterp-action` API to
-peek into the list of actions yet to be performed.
-Once an action is performed, the API reveals the next one. See also the
-related utilities `coq.replay-synterp-action` and
-`coq.replay-all-missing-synterp-actions`.
+Such synterp actions can be recorded into (nested) groups whose structure is
+declared using well-bracketed calls to predicates `coq.begin-synterp-group`
+and `coq.endsynterp-group` in the synterp phase. In the interp phase, one can
```
gares, line 295:
```suggestion
and `coq.end-synterp-group` in the synterp phase. In the interp phase, one can
```

README.md
```diff
@@ -290,11 +290,16 @@ The synterp-command can output data of that type, but also any other data it
...
+then use predicate `coq.replay-synterp-action-group` to replay all the synterp
+actions of the group with the given name at once.
+
+In the case where one wishes to interleave code between the actions of a given
+group, it is also possible to match the synterp group structure at interp, via
+`coq.begin-synterp-group` and `coq.endsynterp-group`. Individual actions that
```
gares, line 301:
```suggestion
`coq.begin-synterp-group` and `coq.end-synterp-group`. Individual actions that
```

coq-builtin.elpi
```diff
@@ -1802,9 +1802,20 @@ type export-module modpath -> synterp-action.
 % execution (aka interp). See also coq.replay-synterp-action
 external pred coq.next-synterp-action o:synterp-action.
 
-% [coq.replay-all-missing-synterp-actions] Execute all actions needed in
-% order to match the actions performed at parsing time (aka synterp)
-external pred coq.replay-all-missing-synterp-actions .
```
gares, line 1807: Question, is this API still implementable?

gares: I mean, can we have a replay-missing-from-current-group and have a outermost group "all"?

gares:
I believe the API is much better than before, I added a few comments.

coq-builtin.elpi
rlepigre: Yes, but it is not a great API since it pierces through abstractions, so we thought that it was better to remove it. The reason why we had to implement this new API is to avoid this precise problem in some of our complex elpi commands that rely on NES internally.

rlepigre: Yes to your second question.

src/coq_elpi_builtins_synterp.ml
rlepigre: Done.

coq-builtin.elpi
rlepigre: Actually I don't know about group "all" in a context where you compose commands, it is not clear who would begin/end it. We can easily add a `replay-missing-from-current-group` though.

gares: Doesn't  coq.next-synterp-action pierces it? (it is just above)

Shoudn't you add two constructors to the synterp-action to be used when the next action is open/closing a group? In a way,coq.replay-all-missing-synterp-actions was also implementable via coq.next-synterp-action + an utility in coq-lib to replay one action (mentioned in the comment above).

So either you remove the coq.next-synterp-action (which I'm not very sure about) or make it return the open/closing bracket. IMO the latter would be a good compromise, since one can implement in pure elpi an a function that, for example, closes the current box, and go back using the high level APIs. I'm sure I use next-synterp-action in HB for example, not sure removing it is worth it.

rlepigre: > Doesn't coq.next-synterp-action pierces it? (it is just above)

It fails if the group has not been opened, so no. At interp, you need to either use `replay-synterp-action-group "my-group"` or explicitly do `begin-synterp-group "my-group" G`, a bunch of `coq.next-synterp-action`, and `end-synterp-group G`. Potentially with more nesting if you have sub-groups.

coq-builtin.elpi
```diff
@@ -1797,14 +1797,29 @@ type include-module-type modtypath -> synterp-action.
 type import-module modpath -> synterp-action.
 type export-module modpath -> synterp-action.
 
+% Synterp action group
+kind group type.
+type grp int -> string -> group.
```
Janno, line 1802: Should this be a typeabbrev to some `ctype "group"` without any constructors?

src/coq_elpi_builtins_synterp.ml
```diff
@@ -439,11 +629,22 @@ module SynterpAction = struct
...
+      | Ok (group, rz) ->
+        List.iter (fun {action; resulting_state} ->
+            Vernacstate.Synterp.unfreeze resulting_state;
+            ignore @@ replay1 action state
+          ) group;
+        rz
+      | Error `Empty ->
+        synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to replay an action group (" ^ name ^ ") in the interp phase but there are no remaining synterp actions or groups.")))
+      | Error `LeadingAction ->
+        synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to replay an action group (" ^ name ^ ") in the interp phase but the next synterp actions is an action and not a group.")))
```
Janno, line 644:
```suggestion
        synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to replay an action group (" ^ name ^ ") in the interp phase but the next synterp action is an action and not a group.")))
```

src/coq_elpi_builtins_synterp.ml
```diff
@@ -607,24 +808,60 @@ module SynterpAction = struct
...
+
+    MLCode(Pred("coq.end-synterp-group",
+      In(Tree.group, "Group",
+      Full(unit_ctx,"Match a the end-synterp-group synterp operation. Group must be the currently opened synterp action group and the group must not have any more synterp actions or groups left to replay.")),
+    (fun name ~depth _ _ state ->
+      State.update read state (fun rz ->
+        match RZipper.close_group rz name with
+        | Ok wz -> wz
+        | Error `NoGroup ->
+          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match a end-synterp-group operation on group " ^ Tree.group_name name ^ " but there is no group to end.")))
```
Janno, line 858:
```suggestion
          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match an end-synterp-group operation on group " ^ Tree.group_name name ^ " but there is no group to end.")))
```

src/coq_elpi_builtins_synterp.ml
```diff
@@ -607,24 +808,60 @@ module SynterpAction = struct
...
+        | Error (`GroupMismatch name') ->
+          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match a end-synterp-group operation on group " ^ Tree.group_name name ^ " but the next group to close has name " ^ Tree.group_name name' ^ ".")))
```
Janno, line 860:
```suggestion
          synterp_interp_error Pp.(hov 0 (strbrk ("The command tried to match an end-synterp-group operation on group " ^ Tree.group_name name ^ " but the next group to close has name " ^ Tree.group_name name' ^ ".")))
```

coq-builtin.elpi
gares: Absolutely,the type should be opaque. See the OpaqueData module in the API

rlepigre: Done.

coq-builtin.elpi
Janno: Here's what we could easily do in elpi itself: we made `coq.next-synterp-action` fail softly when there aren't any more actions so we could call that repeatedly until it fails and just replay the actions it returns. It would work at the top level just like the old primitive worked and it would also work to finish all actions in the current group but it wouldn't silently pierce anybody's group abstractions.

gares: What you propose seems sensible, let me just double check I understood.

It seems I could take any piece of code replaying some actions via the next-synterp-action and call it from a context that puts it in a box (a group) and the code should just work. Did I get it right? If so, 100% OK.

Janno:
@gares do you see anything wrong with the way that elpi state (the evarmap in particular) is handled by our changes? In particular, could any of the primitives we define confuse synterp and interp evarmap state (if there even is such a thing as synterp evarmap state)?

We are currently debugging very confusing failures that seem to be caused by running the interp phase of `nes.end-path`. https://coq.zulipchat.com/#narrow/stream/237656-Coq-devs-.26-plugin-devs/topic/When.20does.20.60new_type_evar.60.20return.20an.20instantiated.20evar.3F/near/424662457 is one of the symptoms but we can produce lots of other assertions failures and also loops in ` elpi_solution_to_coq_solution`, all pointing to evarmap invariants being broken.

We can't say for sure that these problems wouldn't happen without this change here since without it we never managed to port our elpi code to the point where these mysterious errors are popping up. We haven't been able to reproduce the problem with a simple test case, unfortunately.

coq-builtin.elpi
rlepigre: Yes, that's exactly right.

rlepigre: This is implemented now.

rlepigre:
@gares As far as we're concerned, this is ready (we've managed to fix all our elpi commands using this MR).

gares:
Great news!

gares (approved):

PR #605, opened 2024-02-29
Fix a couple of bugs.

rlepigre:
CC @Janno.

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -1855,7 +1855,12 @@ Supported attributes:
      state, !: s, [])),
   DocAbove);
 
-  Coq_elpi_builtins_synterp.current_path;
+  MLCode(Pred("coq.env.current-path",
+    Out(list B.string, "Path",
+    Read(unit_ctx, "lists the current module path")),
+  (fun _ ~depth _ _ state -> !: (mp2path (Global.current_modpath ())))),
+  DocAbove);
+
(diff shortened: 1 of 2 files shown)


src/coq_elpi_builtins.ml
```diff
@@ -1855,7 +1855,13 @@ Supported attributes:
...
   DocAbove);
 
-  Coq_elpi_builtins_synterp.current_path;
+  MLCode(Pred("coq.env.current-path",
+    Out(list B.string, "Path",
+    Read(unit_ctx, "lists the current module path")),
+  (fun _ ~depth _ _ state -> !: (mp2path (Global.current_modpath ())))),
+  DocAbove);
+
+  (* FIXME should be different from the Synterp version. *)
```
gares, line 1864: ?

gares (approved):
I'd merge, but what is this comment?
Do you want to add an assert?

gares:
Thanks!

src/coq_elpi_builtins.ml
rlepigre: Basically, the following builtin has the same problem, but it cannot be fixed as the information is not available at the `interp` phase. There are several options: removing the comment and ignoring this, leaving the comment to remind us that there are problems with this primitive, removing the primitive (only at `interp`).

gares: I think an issue and a pointer to that issue in the doc of the API would be good

gares (approved):

PR #608, opened 2024-03-05
fix uvamp synchronization

gares:
This should fix the problem we faced today (the loop and the Not_found). Please test.

CC @janno @rlepigre


rlepigre:
This did help with the loop and the `Not_found`, but these were caused by a state bug in #604 (now fixed). That being said, this MR could still be useful regardless, if only to help debug such bugs in the future.

gares:
OK good, but I need to "promote" the changes in the doc...

PR #612, opened 2024-03-18
stash conv problems before typechecking

Tragicus:
If we enter `coq.typecheck` or `coq.typecheck-ty` with conversion problems in the evar map, the plugin will try to solve them immediately after typechecking, which is incorrect.


gares:
Thanks. Can I also have a test? From the description of the PR is not immediately clear to me how to take advantage of the change. Also, it seems to suggest an API to force the resolution of these delayed constraints is missing.

PR #614, opened 2024-03-19
Fix roots in solution2evd

Tragicus:
It may happen that coq-elpi aliases something in the input evar_map and then drops the alias because it does not appear in the solution returned by coq-elpi. However, this alias might be referenced in the evar_map, so it should not be dropped. This PR fixes the issue by adding the evars from the input evar_map to the roots of the computation of reachable evars.

### DIFF
--- a/src/coq_elpi_HOAS.ml
+++ b/src/coq_elpi_HOAS.ml
@@ -2462,7 +2462,7 @@ let reachable sigma roots acc =
 let solution2evd sigma0 { API.Data.constraints; assignments; state; pp_ctx } roots =
   let state, solved_goals, _, _gls = elpi_solution_to_coq_solution ~calldepth:0 constraints state in
   let sigma = get_sigma state in
-  let roots = Evd.fold_undefined (fun k _ acc -> Evar.Set.add k acc) sigma0 roots in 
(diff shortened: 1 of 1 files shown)


CohenCyril:
@Tragicus could you link to a branch of Coq which exposes the primitive you needed?

Tragicus:
Yes, here it is (I only need the contents of `engine/evd.ml`): https://github.com/Tragicus/coq/tree/utils
For reference, I opened a PR for this branch: https://github.com/coq/coq/pull/18820

gares:
I think the old code was actually implementing Evd.undefined_evars, about the other one I don't know.

gares (approved):

gares:
@Tragicus can you tell me if this code does what you wanted?

Then we definitely need to do some benchmarking

Tragicus:
It was long ago so I might be mistaken, but I am pretty sure this solved our issue back then. I can do the benchmarking tomorrow.

Tragicus:
I can compile neither `master` nor `coq-master` neither on coq's `master` nor `v8.19`. Can you tell me which branches I should take?

gares:
Maybe you need to rebate this pr, but currently master works with 8.19, 8.20rc1 and master, both with opam and Nix.

PR #615, opened 2024-03-20
Port to the [dune] build system.

rlepigre:
This goes a bit further than #567, but there are still things to figure out:
- [ ] Making sure CI still works.
- [ ] Figure out whether there are requirements for Coq CI.
- [ ] Figure out how to turn more tests into cram tests (to get more silent build).
- [x] Figure out how to explicitly tell dune about the generated `coq-builtin[-synterp].elpi` files so that we can promote.
- [x] Fix doc target


apps/derive/elpi/dune
```diff
@@ -0,0 +1,18 @@
...
+  (glob_files *.elpi))
+ (action
+  (with-stdout-to %{target}
+   (echo ""))))
+
+(install
+ (files
+  (glob_files (*.elpi with_prefix coq/user-contrib/elpi/apps/derive/elpi/)))
+ (section lib_root)
+ (package coq-elpi))
```
rlepigre, line 18: This setup seems to work pretty well, and takes care of installing `.elpi` files.

The only drawback is that we need to put these files under a different path (here files are found under `elpi.apps.derive.elpi`, while they were under `elpi.apps.derive` before). That's probably not a big deal?

builtin-doc/dune
```diff
@@ -0,0 +1,12 @@
...
+ (libraries elpi_plugin))
+
+(rule
+ (targets
+  coq-builtin.elpi
+  coq-builtin-synterp.elpi
+  elpi-builtin.elpi)
+ (deps gen_doc.exe)
+ (mode promote)
+ (action (run ./gen_doc.exe)))
```
rlepigre, line 12: With this setup, the `Elpi Document Builtins` is probably useless?

gares:
I'll rebase and try to fix the doc target, which is a blocker

.nix/coq-overlays/coq-elpi/default.nix
```diff
@@ -61,10 +61,9 @@ in mkCoqDerivation {
   release."1.6.0".sha256      = "0kf99i43mlf750fr7fric764mm495a53mg5kahnbp6zcjcxxrm0b";
   releaseRev = v: "v${v}";
 
-  buildFlags = [ "OCAMLWARN=" ];
-
   mlPlugin = true;
-  propagatedBuildInputs = [ coq.ocamlPackages.findlib elpi ];
+  useDune = true;
+  propagatedBuildInputs = [ coq.ocamlPackages.dune_3 elpi coq.ocamlPackages.ppx_optcomp ];
```
gares, line 66:
```suggestion
  propagatedBuildInputs = [ coq.ocamlPackages.dune_3 coq.ocamlPackages.findlib  elpi coq.ocamlPackages.ppx_optcomp ];
```

PR #633, opened 2024-05-31
Fast check for the presence of a universe in the environment.

ppedrot:
The previous code was very inefficient, as it was basically rebuilding the whole set of universes from the environment in O(n) at each invocation.

### DIFF
--- a/src/coq_elpi_builtins.ml
+++ b/src/coq_elpi_builtins.ml
@@ -508,12 +508,19 @@ let set_accumulate_to_db_interp, get_accumulate_to_db_interp =
   (fun x -> f := x),
   (fun () -> !f)
 
+let is_global_level env u =
+  try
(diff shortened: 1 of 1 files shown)


gares:
Thanks! May I ask how you found this inefficiency? Is there a development suffering from this problem in CI?

ppedrot:
Random runtime profiling of mathcomp while I was debugging something else. FTR there are *a lot* of low-hanging fruits in coq-elpi, the leading one being probably the little dance to keep global and local constraints in sync, which accounts for 1/3 of the coq-elpi runtime or so on average.

gares:
Yup. You did tell me in some API it should be possible to just thread this data instead of recomputing it, but I naver managed to find the time to try to kill the `demote` crap... help is mostly welcome.

PR #645, opened 2024-06-27
Better elpitime for benching

FissoreD:
elpitime (if active) also prints the time of get_and-compile 

run_static_check try/catch error so that elpitime message can be displayed. 
the elpitime message will tell if the execution was successful or not

### DIFF
--- a/src/coq_elpi_vernacular.ml
+++ b/src/coq_elpi_vernacular.ml
@@ -233,8 +233,9 @@ let run ~static_check program query =
     print_debug false;
     rc
   with e ->
+    let e = Exninfo.capture e in
     print_debug true;
-    raise e
+    Exninfo.iraise e
 ;;
 
 let elpi_fails program_name =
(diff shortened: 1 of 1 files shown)


src/coq_elpi_vernacular.ml
```diff
@@ -218,16 +221,20 @@ let run ~static_check program query =
...
+    Coq_elpi_utils.debug Pp.(fun () -> str txt);
+    Coq_elpi_utils.elpitime Pp.(fun () -> str txt)
+  in
+  try
+    let rc = API.Execute.once ~max_steps:!max_steps exe in
+    print_debug false;
+    rc
+  with e ->
+    print_debug true;
+    raise e
```
gares, line 237:
```suggestion
    print_debug true;
    raise e
```
```suggestion
    let e = Exninfo.capture e in
    print_debug true;
    Exninfo.iraise e
```

PR #647, opened 2024-07-02
[derive] Open scope only locally

eponier:
Avoids messing with the scopes opened by the client.

```
Require Import ZArith.
From elpi.apps Require Import derive.std.

Check forall p, p < p.
```
Before
```
forall p : positive, p < p
     : Prop
```
After
```
forall p : nat, p < p
     : Prop
```


gares (approved):
thanks!

PR #648, opened 2024-07-02
[TC] CS projections replaced with uvar

FissoreD:
If a subterm is a (primitive) projection, it is replaced with fresh elpi variable
This variable and the original subterm are related by the coq.unify-eq predicate, to ensure
that each subterm of a class call correctly unifies with the corresponding subterm in the instance.

[Here](https://github.com/FissoreD/coq-elpi/blob/d67f0871e9b32ada1103d93d475de63f2402e0aa/apps/tc/tests/canonical_struct.v) the test for this PR


FissoreD:
Note that the compilation with uvar for projections in this PR only addresses instances compilation

gares:
I'm a bit lost, the test passes but I don't see how the new term constructor is compiled... miracle? 😀

FissoreD:
The precompilator transforms a term like `app[X|T]` (when X is a proj) into `tc.canonical-projection (app [X|T]) S N` (S is the list of free vars in `app[X|T]` and `N` the arg number of the record).  
[Here](https://github.com/LPCIC/coq-elpi/pull/648/files#diff-651e2ec4d32d63306133f396a73dcb6f0aacc0d61bd164ace4c00a3c52611bffR25) I recognize a proj and build the new subterm `tc.canonical-projection`. Note that I also output `+1` for the number of uvar the rule I'm building needs (the last parameter of the clause head is: `(s A)`). 

---

[Here](https://github.com/LPCIC/coq-elpi/pull/648/files#diff-651e2ec4d32d63306133f396a73dcb6f0aacc0d61bd164ace4c00a3c52611bffR25) I decompile tc.canonical-projection (during the elpi's rule creation): I replace the sub-term with a fresh elpi-var `V` (that I have charge with pi earlier) and create the new premise `coq.unify-eq (app[X|T]) V ok`.

---

Am I clarifying your remark ?

gares (approved):

gares:
@CohenCyril  this job https://github.com/LPCIC/coq-elpi/actions/runs/11143050022/job/30967371306?pr=648
indicated that I'm using elpi 1.19.2 but in the nix config i chose 1.19.5. What am I missing?

FissoreD:
I think this can be closed in favour of #1045

PR #656, opened 2024-07-08
[TC] premise not run if rigid solution

FissoreD:


apps/tc/elpi/tc_aux.elpi
```diff
@@ -109,19 +109,20 @@ pred get-mode i:gref, o:list string.
...
 
-pred make-tc.aux i:bool, i:prop, i:list prop, o:prop.
-make-tc.aux _ Head [] Head.
-make-tc.aux tt Head Body (Head :- Body).
-make-tc.aux ff Head Body (Body => Head).
+pred make-tc.aux i:bool, i:term, i:prop, i:list prop, o:prop.
+make-tc.aux tt _ Head [] Head :- !.
+make-tc.aux ff Sol Head [] P :- !, P = if (var Sol) Head true.
+make-tc.aux tt _ Head Body (Head :- Body) :- !.
+make-tc.aux ff Sol Head Body P :- !, P = if (var Sol) (Body => Head) true.
```
gares, line 116: Should it be `ground_term`? It may not be super efficient, so we can leave it like that but at least put a comment that the right test should be `ground_term`

FissoreD: Absolutely. 
At first my code used ground_term but then I was doubting about performance issues. 
That's why I put a simple var check
I'll put the comment

gares (approved):

PR #666, opened 2024-07-16
Adapt to coq/coq#19361 (Lib.*.close_section is incomplete, use Declaremods instead)

SkySkimmer:


SkySkimmer:
The elpi tests about sections may be incomplete, https://gitlab.inria.fr/coq/coq/-/jobs/4563952 only caught the issue because we rebuilt the plugin with warnings as errors (hb tests did catch the issue with real tests https://gitlab.inria.fr/coq/coq/-/jobs/4563953)

gares (approved):

PR #671, opened 2024-07-19
Do not escape quotes in verbatim LPDoc documentation.

ppedrot:
This confuses my text editor and I think this is actually useless. Indeed, looking at the code of Elpi seems to indicate that we do not need one additional level of escaping there.

@gares please check that this analysis is correct.


gares (approved):
thanks

PR #673, opened 2024-07-21
Adapt to Coq PR #19404: an algebra of types for the instances of notation variables

herbelin:
This PR is in preparation of coq/coq#19404 which introduces an algebra of types for the instances of notation variables.

To be merged synchronously.

### DIFF
--- a/src/rocq_elpi_builtins.ml
+++ b/src/rocq_elpi_builtins.ml
@@ -1995,6 +1995,28 @@ let mis_is_recursive { Declarations.mind_recargs } =
 let mis_is_recursive = Inductiveops.mis_is_recursive
 [%%endif]
 
+[%%if coq = "9.0" || coq = "9.1" || coq = "9.2"]
+let empty_notation_interp_env = Notation_term.{
+  ninterp_var_type = Id.Map.empty;
+  ninterp_rec_vars = Id.Map.empty
+}
+[%%else]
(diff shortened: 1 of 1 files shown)


gares:
Please use optcomp in order to make the pr work on 8.19 and 8.20

gares:
I'm sorry but the patch is still not acceptable since it only works with master. Here an example of ppx_optcomp

https://github.com/herbelin/coq-elpi/blob/aba88b8af2b5446e40d89933d361273d916fb1b0/src/rocq_elpi_builtins.ml#L155-L167

please put the old code into a function so that you can conditionally compile that (on older versions of rocq) or the new code on master

herbelin:
Not sure I did it correctly (some tests are apparently still failing). I would not mind a bit of help. Thanks in advance.

gares:
The if condition should be a disjunction with 9.0, 9.1 and 9.2, and in the else branch you put the new code. There is an example of the right syntax just above.

herbelin:
Ah, ok. I was pretty naive.

There are still failures. Is there still something wrong on the PR side?

gares:
These are fine, the master branch is tested by rocq ci

gares (approved):

SkySkimmer:
The analysis failures aren't on master

gares:
They seem nix related. You could try rebase or run make nix to update the nix pin

PR #683, opened 2024-09-02
Overlay for PR 19473

mattam82:
To merge when coq/coq#19473 lands.


gares:
Sorry but you need to use optcomp. IIRC there is a syntax for record fields, if not put the entire code in a big ifdef.

PR #693, opened 2024-09-20
ifdefs on elpi version in source code

gares:


gares:
@vbgl I need some nix expertise here.
It seems that `%{version:elpi}` does not work on nix, it seems to expand to the empty string.
Is it expected? It should be the version written in the META file, in turn generated by dune...

vbgl:
I think the problem comes from the fact that `elpi` in nixpkgs is built from the git repository rather than from the released tarball (which contains correct version information, I guess when produced by dune-release).

gares:
Hem, ok, thanks for explaining.

Next question is: how can I fix that?

Is using the tarball generated by dune-release a problem for nix?

vbgl:
I’ll propose a patch for nixpkgs asap. What is suppose to happen to your dune scripts (in coq-elpi) when using a development version (i.e., not a release with a proper version string) of elpi?

etc/version_parser.ml
```diff
@@ -0,0 +1,17 @@
...
+
+let main () =
+  let v = Sys.argv.(1) in
+  let l = String.split_on_char '.' Str.(replace_first (regexp "^v") "" v) in
+  (* sanitization *)
+  let l =
+    match l with
+    | l when List.for_all is_number l -> l
+    | ["%%VERSION_NUM%%"] -> ["99";"99";"99"]
+    | _ -> Printf.eprintf "version_parser: cannot parse: %s\n" v; exit 1 in
```
gares, line 12: @vbgl this was my "attempt" to support development versions. I'm happy to improve on that

etc/version_parser.ml
```diff
@@ -0,0 +1,17 @@
...
+let is_number x = try let _ = int_of_string x in true with _ -> false ;;
+
+let main () =
+  let v = Sys.argv.(1) in
+  let l = String.split_on_char '.' Str.(replace_first (regexp "^v") "" v) in
+  (* sanitization *)
+  let l =
+    match l with
+    | l when List.for_all is_number l -> l
+    | ["%%VERSION_NUM%%"] -> ["99";"99";"99"]
```
gares, line 11:
```suggestion
    | ( [""] | ["%%VERSION_NUM%%"] ) -> ["99";"99";"99"]
```

vbgl:
The fix has been merged into nixpkgs master branch. I hope this addresses you problems. Please ping again if it’s not the case. Kudos to Cyril for the patch.

gares:
@CohenCyril opam works on master, but not nix. Since I can't run nix on my machine, is there any chance to get verbose logs in CI?

gares:
I mean, the CI file is generated, and it contains `2>&1 > /dev/null`. Can I disable that "feature"?

PR #694, opened 2024-09-24
Elpi Compile to fill the cache

gares:
@FissoreD you should try placing this at the beginning of your stdpp/iris files so to eliminate the compilation time from your bench.


gares:
This should be used atop of the new elpi, but for that we need #693

PR #703, opened 2024-10-24
Adapt to Coq PR #19301 which unifies the syntax of Theorem, Definition and Fixpoint

herbelin:
The change seems straightforward. I hope I did not break anything though.

To be merged synchronously with coq/coq#19301 when the latter is ready.


gares:
Coq-Elpi master supports two version of Coq (8.20 and master) at the same time thanks to ppx_optcomp.
You broke 8.20 since you did not add appropriate ifdefs.
See this PR for example: https://github.com/LPCIC/coq-elpi/pull/696/files

PR #709, opened 2024-10-28
[tc] attribute parser for the creation of elpi predicates for tc

FissoreD:


FissoreD:
The CI does not pass since it uses elpi-1.19.2 and in the PR there is a call to std.once added in [elpi-1.19.4](https://github.com/LPCIC/elpi/releases/tag/v1.19.4)

PR #714, opened 2024-11-08
API for open (as in free variables) terms

gares:


examples/extensional_replace.v
```diff
@@ -0,0 +1,448 @@
...
+
+mk-equality RW (app L as C) A C {{@refl_equal _ lp:C}} A :-
+fold-map2 L A (mk-equality RW) _ L1 _A1,
+std.forall L1 (c \ (sigma A B \ c = {{@refl_equal lp:A lp:B}})),!
+, coq.say "tenth clause"
+.
+
+mk-equality RW {{@map lp:T1 lp:T2 lp:{{fun N _ F}} lp:L}}
+  A
+  {{@map lp:T1 lp:T2 lp:{{fun N T1 F1}} lp:L1}}
```
gares, line 239: @ybertot this rule uses `N` all over the place in the resulting term/proof. I suspect this is where the collision  is perpetuated.

While I'm thinking at it maybe there is a simple alternative: Start by calling copy on the statement with a rule that enforces the "Barendregt convention" using the fresh-id API, and then simply run your code (and make no more a call to fresh-id in remove-one-unknown).
If I'm right that the API is really the same Coq's PP uses, then it should work since the data the code runs on would really reflect what the user sees when he inputs the replacement command. In a way we repair, upfront, the discrepancy the PP introduces.

ybertot:
Yes, this is what I wanted to tell you orally.

gares:
in favor of https://github.com/LPCIC/coq-elpi/pull/720

Issue #716, opened 2024-11-22
`coq.TC.get-inst-prio` reports wrong priority


Janno:
```coq
Require Import elpi.elpi.

Class Test : Prop := {}.

Instance instance_c : Test := Build_Test.
Instance instance_g : Test | 4 := Build_Test.

Definition hint_c : Test := Build_Test.
Definition hint_g : Test := Build_Test.

Hint Resolve hint_c : typeclass_instances.
Hint Resolve hint_g | 5 : typeclass_instances.

Elpi Command test.
Elpi Accumulate lp:{{
  pred show i:gref.
  show Class :- std.forall {coq.TC.db-for Class} (show.inst Class).
  pred show.inst i:gref, i:tc-instance.
  show.inst Class (tc-instance Instance P1) :-
    coq.TC.get-inst-prio Class Instance P2,
    coq.say Instance "," P1 "," P2.
}}.
Elpi Query lp:{{ show {{:gref Test}} }}.
(*
Output:
<<
const «hint_c» , tc-priority-computed 0 , tc-priority-computed 0
const «instance_c» , tc-priority-computed 0 , tc-priority-computed 0
const «instance_g» , tc-priority-given 4 , tc-priority-given 4
const «hint_g» , tc-priority-computed 0 , tc-priority-computed 0
>>

BUG: <<hint_g>> should have <<tc-priority-given 5>>.
```

Found by @swasey

gares:
CC @FissoreD

gares:
I guess internally the two priorities are not stored in the same place. Is this expected @mattam82 ?

FissoreD:
Mh, I think I've found where to get the prio.

If I understand correctly, when we declare an Instance with a user-defined prio $P$, $P$ is put in the hint_priority field of the instance.

https://github.com/coq/coq/blob/303a7c24ddd383d894b90b98178e2e8ee89f675a/pretyping/typeclasses.ml#L23C1-L26C1

If the prio is not set, then the same field is set to None, and it is in that situation that we try to figure out the approximated priority of the instance by counting the number of its hypothesis.

https://github.com/LPCIC/coq-elpi/blob/11a121e2e20013d01e0c70ce2d3b737415c0c624/src/coq_elpi_builtins.ml#L445-L464

Apparently the instruction:
`Hint Resolve hint_g | 5 : typeclass_instances.`
adds the instance but put None into the field hint_priority.

[...]

PR #717, opened 2024-11-25
[coq.TC.get-inst-prio] new get-inst-prio (#716)

FissoreD:


FissoreD:
With the new push, I'm not sure the tc-priority remains relevant, orelse, if we want to keep this piece of information we need to reuse the old algorithm to know if the instance has a UserGiven priority.

Note that in that case, I'm not sure how to know if a hint (added with hint resolve) has a UserGiven priority

gares:
In these cases I think having the exception carry the found value is less cumbersome.

Otherwise looks good.

gares (approved):

PR #725, opened 2024-12-02
nix

gares:


gares:
@CohenCyril help plz. Isn't this the way you override the version of elpi?
```
  bundles = {

    "coq-8.20" = {
      coqPackages = common-bundles // {
        coq.override.version = "8.20";
      };
      ocamlPackages = {
        elpi.override.version = "v2.0.3";
      };
    };
```

CohenCyril:
@gares for some strange reason that I cannot recall the version of elpi at use is locally overriden in the definition of the coq-elpi derivation... AFACT this looks like bad practice. Perhaps the reason is that when we override coq-elpi there is a high chance that we want to override elpi at the same time and this does the job in a confusing way.
Anyway I will push a patch asap and later this week try to fix it upstream.

CohenCyril:
```
       error: ppxlib-0.15.0 is not available for OCaml 4.14.2
```
what version of ocaml is appropriate?

gares:
That version is fine, 4.13 is the minimum

gares:
Who fixés ppxlib?

PR #737, opened 2025-01-06
Add `coq.env.add-context` for inserting context declarations

VojtechStep:
There was no way to add implicit section variables, so `context-decl` objects could not be faithfully spliced from outside code.

I'd appreciate a naming suggestion for what I temporarily called `coq.env.add-section-variable-two`.


gares (approved):
Thanks, it looks good.

gares:
I would be tempted to break the existing api, giving a default value to the argument when omitted.

Surely HB will need adjustment.
Would you mind trying and see what CI does discover?

VojtechStep:
> Would you mind trying and see what CI does discover?

I can try tomorrow.

gares:
Also, I forgot: here an example of an argument with a default:
https://github.com/LPCIC/coq-elpi/blob/ee28024f5e10c25fe5e102ae553d2d4ef0e42a1a/src/coq_elpi_builtins.ml#L1870-L1878

CohenCyril:
> I would be tempted to break the existing api, giving a default value to the argument when omitted.

How about using an optional `@implicit! => coq.env.add-section-variable Name Ty C` to preserve backward compat?

VojtechStep:
No worries, I figured it out in the meantime 👌

gares:
It seems HB is the only package impacted by the change, so I think this patch is fine.
I'll try to see how to make HB compile with 2.3.0 and also with this change.

gares:
waiting https://github.com/ocaml/opam-repository/pull/27258

gares:
I tried to make nix build against the upstream PR and tag but I failed.
I guess I'll wait for a proper release...

CohenCyril:
> @CohenCyril I need help to override the version of elpi (2.0.7 is in nixpkgs already)

I don't have push rights to @VojtechStep

CohenCyril:
The fix is [here](https://github.com/VojtechStep/coq-elpi/pull/1)

CohenCyril:
The error on master is related to the Coq -> Rocq renaming, which I'm having a hard time to follow these days

.nix/config.nix
```diff
@@ -14,6 +14,7 @@ let master = [
   common-bundles = listToAttrs (forEach master (p:
     { name = p; value.override.version = "master"; }))
   // {
+    hierarchy-builder.override.version = "VojtechStep:feature/breaking-add-section-variable";
```
gares, line 17:
```suggestion
```

.nix/coq-overlays/coq-elpi/default.nix
```diff
@@ -81,7 +81,7 @@ in
   defaultVersion = lib.switch coq.coq-version [
     {
       case = "8.20";
-      out = "2.2.0";
+      out = "#737";
```
gares, line 84:
```suggestion
      out = "2.2.0";
```

Changelog.md
```diff
@@ -1,3 +1,11 @@
+# UNRELEASED
```
gares, line 1:
```suggestion
# [2.4.0] 15/1/2025
```

PR #740, opened 2025-01-15
Adapt w.r.t. coq/coq#20060.

ppedrot:


ppedrot:
This should actually be backwards compatible, so I'm undrafting.

gares (approved):

PR #742, opened 2025-01-17
Adapt to coq/coq#19985 (template poly has pseudo sort poly)

SkySkimmer:


SkySkimmer:
Changed upstream so this isn't needed

PR #747, opened 2025-01-19
apply tactic, similar to rewrite

patrick-nicodemus:
Let me know if you can suggest useful ways to eliminate duplication here between apply and rewrite, both of them are doing the same thing where they traverse nested foralls.

Is repeatedly calling `refine` and failing inefficient? I have no idea.

I would also be interested if you have suggestions for making it stronger, it would be nice to have it behave as 
"if the type of term `f` is *unifiable* with `forall (x : A), B`, then recursively apply `f _`. But I don't know how to check that checks whether a type is unifiable with a product type.

### DIFF
--- a/apps/eltac/theories/apply.v
+++ b/apps/eltac/theories/apply.v
@@ -7,7 +7,6 @@ Elpi Accumulate lp:{{
     apply T _ G GL :- refine T G GL.
 
     apply Term (prod _ A B) G GL :-
-        Hole = {{ _ }},
         coq.typecheck Hole A ok,
         whd (app [Term, Hole]) [] HD ARGS,
         unwind HD ARGS Term',
@@ -16,9 +15,7 @@ Elpi Accumulate lp:{{
     apply _ _ _ _ :- coq.ltac.fail _ "Couldn't unify type of term with goal".
         
     solve (goal Ctx _ _ _ [trm T] as G) GL :-
-        (% T is a direct reference to a global definition or axiom
-        T = global Gref, coq.env.typeof Gref Ty;
-        % T is a direct Gallina term and we will infer its type
+        (% T is a direct Gallina term and we will infer its type
         % from context
         coq.typecheck T Ty ok;
         % Eq is a reference to a declared variable in the context
--- a/apps/eltac/theories/apply.v
+++ b/apps/eltac/theories/apply.v
@@ -6,8 +6,7 @@ Elpi Accumulate lp:{{
 
     apply T _ G GL :- refine T G GL.
 
-    apply Term (prod _ A B) G GL :-
-        coq.typecheck Hole A ok,
+    apply Term (prod _ _ B) G GL :-
         whd (app [Term, Hole]) [] HD ARGS,
         unwind HD ARGS Term',
         apply Term' (B Hole) G GL.
--- a/apps/eltac/theories/apply.v
+++ b/apps/eltac/theories/apply.v
@@ -15,9 +15,7 @@ Elpi Accumulate lp:{{
     solve (goal Ctx _ _ _ [trm T] as G) GL :-
         (% T is a direct Gallina term and we will infer its type
         % from context
-        coq.typecheck T Ty ok;
-        % Eq is a reference to a declared variable in the context
-        std.mem Ctx (decl T _ Ty)),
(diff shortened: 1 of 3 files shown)


patrick-nicodemus:
Not sure if coq.saturate is appropriate here, if we want to be able to apply `f : A -> (B -> C)` to the goal `B -> C`.
Unless we want to saturate both the term to be applied and the goal.

apps/eltac/theories/apply.v
```diff
@@ -0,0 +1,26 @@
...
+        apply Term' (B Hole) G GL.
+
+    apply _ _ _ _ :- coq.ltac.fail _ "Couldn't unify type of term with goal".
+        
+    solve (goal Ctx _ _ _ [trm T] as G) GL :-
+        (% T is a direct Gallina term and we will infer its type
+        % from context
+        coq.typecheck T Ty ok;
+        % Eq is a reference to a declared variable in the context
+        std.mem Ctx (decl T _ Ty)),
```
gares, line 21: Why do you want to check this? Can't it be an arbitrary term?

apps/eltac/theories/apply.v
```diff
@@ -0,0 +1,26 @@
...
+Elpi Tactic apply.
+Elpi Accumulate lp:{{
+    pred apply i:term, i:term, o:goal, o:list sealed-goal.
+
+    apply T _ G GL :- refine T G GL.
+
+    apply Term (prod _ _ B) G GL :-
+        whd (app [Term, Hole]) [] HD ARGS,
+        unwind HD ARGS Term',
+        apply Term' (B Hole) G GL.
```
gares, line 12: There seems to be a misconception here, I think you want to reduce the type, not the proof term
Untested:
```suggestion
    apply Term Ty G GL :-
        whd Ty [] (prod _ _ B) [],
        apply {coq.mk-app Term [Hole]} (B Hole) G GL.
```

apps/eltac/theories/apply.v
```diff
@@ -0,0 +1,26 @@
+From elpi Require Export elpi.
+
+Elpi Tactic apply.
+Elpi Accumulate lp:{{
+    pred apply i:term, i:term, o:goal, o:list sealed-goal.
+
+    apply T _ G GL :- refine T G GL.
```
gares, line 7:
```suggestion
    apply T _ G GL :- refine T G GL, !.
```

gares (commented):
Thanks for your contribution, please have a look at my comments.

gares:
> Let me know if you can suggest useful ways to eliminate duplication here between apply and rewrite, both of them are doing the same thing where they traverse nested foralls.
> 
> Is repeatedly calling `refine` and failing inefficient? I have no idea.
> 
> I would also be interested if you have suggestions for making it stronger, it would be nice to have it behave as "if the type of term `f` is _unifiable_ with `forall (x : A), B`, then recursively apply `f _`. But I don't know how to check that checks whether a type is unifiable with a product type.

I did do a `whd`. If you want to be stronger and unify, you can call `coq.unify-eq Ty {{ forall x, lp:(B x) }} ok` (the quotation is not necessary, but looks nicer than `(prod _ _ B)` (or `(prod _ _ x\B x)` that I do recomment since it documents B takes an argument.

apps/eltac/theories/apply.v
patrick-nicodemus: I just want the type `Ty` of the term `T` somehow, so that I can iteratively drop down through the nested foralls of `Ty`.

gares: Then the typecheck line suffices, no need for the '; mem ...' part

apps/eltac/theories/apply.v
```diff
@@ -0,0 +1,23 @@
...
+
+    apply Term Ty G GL :-
+        whd Ty [] (prod _ _ B) [],
+        apply {coq.mk-app Term [Hole]} (B Hole) G GL.
+
+    apply _ _ _ _ :- coq.ltac.fail _ "Couldn't unify type of term with goal".
+
+    solve (goal Ctx _ _ _ [trm T] as G) GL :-
+        (% T is a direct Gallina term and we will infer its type
+        % from context
```
gares, line 17:
```suggestion
```

PR #752, opened 2025-01-23
fix(ci.yml): Use `coqorg/coq:8.20` as a future-proof image name

erikmd:
Given 8.20.0 has been replaced with 8.20.1
while both images share the 8.20 tag shorthand.

Cc @proux01


gares:
Thanks

PR #776, opened 2025-02-18
Make pos_eq_dec transparent

eponier:
PR https://github.com/LPCIC/coq-elpi/pull/767 replaced `Pos.eq_dec` coming from stdlib with `pos_eq_dec` defined locally. But the former is transparent, while the latter is opaque. This makes `eqb_body` not compute, which is painful.


SkySkimmer:
Is there an argument for moving these lemmas to corelib? It seems strange to avoid a dependency on stdlib only to end up copying its lemmas.

eponier:
I find it also a bit sad to duplicate this kind of definition, but I do not have a good idea of what the tradeoffs are.

Note that `bool_dec` is also redefined, and also opaque instead of transparent. But is used only in the proofs, and does not need to be transparent.

gares (approved):

gares:
> Is there an argument for moving these lemmas to corelib?

yes!

PR #781, opened 2025-02-21
Adapt w.r.t. coq/coq#20278.

ppedrot:
This should be backwards compatible.


gares (approved):

ppedrot:
Since you're fine with this, let's merge already.

PR #790, opened 2025-03-04
Declare compatibility constant as canonical gref for primitive projections

Tragicus:
As of now, we can not obtain a gref out of a primitive projection, although the compatibility constant is a canonical gref associated to it. This is useful when we want to turn a term into a cs-pattern (although I guess I could also add a clause to the latter predicate).

### DIFF
--- a/elpi/coq-lib.elpi
+++ b/elpi/coq-lib.elpi
@@ -510,6 +510,7 @@ coq.term->gref (global GR) GR :- !.
 coq.term->gref (pglobal GR _) GR :- !.
 coq.term->gref (app [Hd|_]) GR :- !, coq.term->gref Hd GR.
 coq.term->gref (let _ _ T x\x) GR :- !, coq.term->gref T GR.
+coq.term->gref (primitive (proj Proj _)) GR :- !, primitive-projection? Proj GR.
 :name "term->gref:fail"
 coq.term->gref Term _ :-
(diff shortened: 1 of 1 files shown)


gares:
There is `primitive-projection?` already. The code is a bit different, and the documentation I wrote is not good.
I think it does what you want but only if the projection is also flagged as canonical.

Either you can close this PR and use the existing API (if it works for you) or we find a way to merge the two.

Tragicus:
I missed `primitive-projection?`, thank you. The point is that I want for it to be used in `term->gref`.

elpi/coq-lib.elpi
```diff
@@ -510,6 +510,7 @@ coq.term->gref (global GR) GR :- !.
 coq.term->gref (pglobal GR _) GR :- !.
 coq.term->gref (app [Hd|_]) GR :- !, coq.term->gref Hd GR.
 coq.term->gref (let _ _ T x\x) GR :- !, coq.term->gref T GR.
+coq.term->gref (primitive (proj Proj _)) GR :- !, primitive-projection? Proj GR.
```
gares, line 513:
```suggestion
coq.term->gref (primitive (proj Proj _)) (con C) :- !, coq.env.primitive-projection? Proj C.
```

Tragicus: I am surprised I did not see any error. Is it not `const` instead of `con`?

gares (approved):

PR #791, opened 2025-03-05
Give constant to every primitive projection

Tragicus:
Followup to https://github.com/LPCIC/coq-elpi/pull/790, because I do need the compatibility constant for every primitive projection, not only for canonical ones.
@gares Should I add another predicate?

### DIFF
--- a/src/rocq_elpi_builtins.ml
+++ b/src/rocq_elpi_builtins.ml
@@ -2363,11 +2363,8 @@ denote the same x as before.|};
   MLCode(Pred("coq.env.primitive-projection?",
     In(projection, "Projection",
     Out(constant, "Compatibility constant",
-    Easy "relates a projection to its compatibility constant.")),
-    (fun p _ ~depth ->
-      let c = Projection.constant p in 
-      try !: (ignore (Structures.Structure.find_from_projection c); Constant c)
-      with Not_found -> raise No_clause)),
+    Easy "relates a primitive projection to its compatibility constant.")),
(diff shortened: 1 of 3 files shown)


gares:
Ideally yes and make this one bidirectional. So
- `coq.env.primitive-projection? P C` would succeed only if the projection is primitive and give P if C is provided or the converse is C is provided. See for example the code of projection-unfolded where both args are `InOut`
- `coq.CS.canonical-projection? C` that tells if C is a projection that can be used for CS inference, the code is the one you are removing now, roughly

gares (approved):

src/rocq_elpi_builtins.ml
```diff
@@ -2361,13 +2361,30 @@ denote the same x as before.|};
...
-    Out(constant, "Compatibility constant",
-    Easy "relates a projection to its compatibility constant.")),
-    (fun p _ ~depth ->
-      let c = Projection.constant p in 
-      try !: (ignore (Structures.Structure.find_from_projection c); Constant c)
-      with Not_found -> raise No_clause)),
+    InOut(B.ioarg projection, "Projection",
+    InOut(B.ioarg constant, "Compatibility constant",
+    Out(int, "Index",
+    Read(global, "Relates a primitive projection to its compatibility constant and its index in the record.")))),
```
gares, line 2367:
```suggestion
    Read(global, "Relates a primitive projection to its compatibility constant. Index is set to the constructor argument extracted by the projection (starting from 0).")))),
```

PR #800, opened 2025-04-08
Adapt to rocq-prover/rocq#20483 (Libobject.simple_open is shallow only)

SkySkimmer:
IDK if it really makes sense to do deep imports for ELPI-UNITS so I kept it with unchanged behaviour.


SkySkimmer:
Please merge now

ppedrot (approved):

PR #807, opened 2025-04-20
lib is no more a keyword (fix #794)

### LINKED ISSUE #794: Importing coq-elpi breaks references that start with lib
This is probably because of the lib: syntax. Maybe the grammar can be tweaked to fix the problem?

```
Module lib.
  Definition foo := I.
End lib.
Check lib.foo.                  (* works fine *)
Require Import elpi.elpi.
Check lib.foo.                  (* syntax error *)
```

gares:


src/rocq_elpi_vernacular_syntax.mlg
```diff
@@ -78,6 +78,19 @@ let lib_ref id =
           ++ str " (you may need to require some .v file with \
                   `Register ... as " ++ str id ++ str ".`).")
 
+let lookahead_lib_colon kwstate strm =
+  match Gramlib.LStream.peek_nth kwstate 0 strm with
+  | Tok.IDENT "lib" ->
+      (match Gramlib.LStream.peek_nth kwstate 1 strm with
+      | Tok.KEYWORD ":" -> ()
+      | _ -> raise Gramlib.Stream.Failure)
+  | _ -> raise Gramlib.Stream.Failure
```
SkySkimmer, line 87: why not use the lookahead combinators?

gares: No special reason, I copied some code from SSR

PR #813, opened 2025-04-24
Port to elpi 3.0

FissoreD:


apps/derive/elpi/projK.elpi
```diff
@@ -76,7 +76,7 @@ body-default (prod N T F) J OJ IT Args K (fun N T B) :- !,
     if (J = OJ)
       (Def = [default-output J Mask Args x T])
       (Def = []),
-    Def => body-default (F x) J1 OJ IT {append Args [pr x T]} K (B x).
+    Def ==> true, body-default (F x) J1 OJ IT {append Args [pr x T]} K (B x).
```
FissoreD, line 79: Why the `true`?
```suggestion
    Def ==> body-default (F x) J1 OJ IT {append Args [pr x T]} K (B x).
```

apps/derive/elpi/param1_trivial.elpi
```diff
@@ -123,7 +123,7 @@ main GR Suffix [Clause,param1-trivial-done GR] :- do! [
   coq.env.add-const Name R RT @transparent! Cst,
   Clause  = (param1-trivial-db (global (indt GR)) (global (const Cst))),
   coq.elpi.accumulate _ "derive.param1.trivial.db" (clause _ _ Clause),
-  coq.elpi.accumulate _ "derive.param1.trivial.db" (clause _ _ (param1-trivial-done GR)),
+  if (not (param1-trivial-done GR)) (coq.elpi.accumulate _ "derive.param1.trivial.db" (clause _ _ (param1-trivial-done GR))) true,
```
FissoreD, line 126: This is a hack to avoid double loading of the same rule

apps/derive/elpi/param1_inhab.elpi
```diff
@@ -119,7 +119,7 @@ main GR Suffix [ClauseW, param1-inhab-done GR] :- do! [
 
   ClauseW = (param1-inhab-db (global (indt GR)) (global (const Witness))),
   coq.elpi.accumulate _ "derive.param1.trivial.db" (clause _ _ ClauseW),
-  coq.elpi.accumulate _ "derive.param1.trivial.db" (clause _ _ (param1-inhab-done GR)),
+  if (not (param1-inhab-done GR)) (coq.elpi.accumulate _ "derive.param1.trivial.db" (clause _ _ (param1-inhab-done GR))) true,
```
FissoreD, line 122: This is a hack to avoid double loading of the same rule

CohenCyril:
> Comments are welcome. I guess one can be smart for fields that are imposed by other (dependent types), and maybe support a trailing _ to pad with as many _ as needed.

Great! That's ground zero of a prototype I wanted to do for a while, handling both nested records and telescope in a single "renaming command".

> What does the lean one do?

Less

CohenCyril:
> I wonder if HB could use the same mechanism, but find the target record type (a factory) by looking at the given fields rather than the expected type as I currently do (and allow `instance Ring int := « mulrC := ... »` as per the blueprint, and `#[prove] instance Ring int := « mulrC := Int.mulC; _ »` alike).

Oh yes, I thought that was the reason you did this. My point is that HB should be split in two: generation and instanciation of records via a generalization of what you did + meta program.

gares:
Well, a generic record constructor seems useful even without HB.

PR #816, opened 2025-05-01
Fix param2 non-uniform param bug

CohenCyril:


gares:
A test would be nice

CohenCyril:
> A test would be nice

Indeed!

gares:
apparently there is one for param1, and my fix is wrong.

CohenCyril:
So my test does not test the fix properly??? I just checked it failed for param2 without the fix though :-/

gares:
My fix to param1 is wrong since there is a régression. Your fix is another story, I guess it is correct.

gares:
I reverted my change, so that CI passes. But looks very weird that my patch was wrong for param1, but is right for param2.

Issue #798, opened 2025-04-08
TC not resilient to small changes deep in a term


Tragicus:
If we feed a term to the typeclass solver, which is not exactly of the expected form, then the latter fails. A funny example of this is when a variable appears several times in the term. Real-life examples include issues with primitive projections and their compatibility constants. I expect we would also have issues with convertible chains of coercions.
```coq
From elpi Require Import elpi tc.

Class C (T : Type) : Type := {}.
Axiom t : forall n m : nat, Type.

Instance tC (n : nat) : C (t n n) := Build_C (t n n).
Set Printing All.
Check _ : C (t 0 0).
(* tC O *)
Check _ : C (t 0 (id 0)).
(* ?c *)

```

PR #820, opened 2025-05-14
Compile with let and unify-eq

FissoreD:
To better control unification, we introduce a new link, `link.unif-eq`, which uses Coq's unification mechanism.
To compile using this link, the user must write the instance using a `let` construct.

For example, in #798, the instance `Instance tC (n : nat) : C (t n n)` uses the same variable twice. 
At runtime, the two occurences of `n` may be instantiated to terms that are unifiable in Coq but not in Elpi.

The solution is to modify the instance as follows:
`Instance tC (n : nat) : let n1 := n in C (t n n1)`
so that it is compiled into:

```elpi
tc-C {{t lp:N lp:N1}} {{tC}} :- tc.link.unif-eq N N1.
```

instead of:

```elpi
tc-C {{t lp:N lp:N}} {{tC}}.
```


gares:
@Tragicus @CohenCyril the plan is to also provide the automatic "let-ification" of the conclusion, eg
```
#[index(shallow)] Instance bla ...
```
that does the let-ification and then registers the instance.

FissoreD:
The function automatically "lettifying" a term is `tc.lettify.main T1 T2`

gares:
I think it can stay, it is just in the wrong directory IMO

PR #823, opened 2025-05-19
Lettify master

gares:


gares:
@CohenCyril @Tragicus this code should allow you to use the TC solver as you need, i.e. call Rocq's unification on all arguments. It would be nice to use this new code in your experiment this week, while I'm in Lyon.

PR #826, opened 2025-05-20
Declare functional predicate api

FissoreD:


FissoreD:
I think the same: an API in elpi is surely a good idea!

apps/tc/tests/func/class.v
```diff
@@ -0,0 +1,9 @@
+From elpi.apps Require Import tc.
+
+Elpi TC.deterministic.
+
+Class C (T : nat).
+
+Instance I1 : C 1. Qed.
+
+Instance I2 : C 1. Fail Qed.
```
FissoreD, line 9: I do not understand why the failure is not thrown after the call to Qed.
If however I do a dummy query like:
```
Elpi Query TC.Solver lp:{{
  true.
}}.
```
we get the error.
Moreover, the Loc in the errors are wrong.

at TC.Solver the error is thrown

FissoreD: Ok, the problem comes from the fact that the observer behaind instance declaration performs a query to the database.
This query calls a `coq.elpi.accumulate-clause`.
After query is performed no check is performed on the database to verify the integrity of DetCheck.
A solution would be to add a `get_and_compile` in `Interp.run_program`.

FissoreD: @gares Is it expensive to make such a check at runtime?

PR #839, opened 2025-06-26
Adapt to rocq-prover/rocq#20810.

ppedrot:
This is backwards compatible.


gares:
Thanks @copilot your wisdom has no match!

PR #847, opened 2025-07-03
Adapt to rocq-prover/rocq#20642 (rm SyntacticOrd)

SkySkimmer:


gares:
I think I use the cache only to reduce memory consumption and opportunistic fast equality, so as long you don't make it unsound things will work.

ppedrot:
Should we just merge then?

PR #851, opened 2025-07-10
derive: remove leftover debug bits

eponier:
#848 introduced some `spy` that I guess should not have landed in master.


gares:
thanks

PR #856, opened 2025-07-22
rbuild: deep fields

gares:


apps/rbuild/examples/usage_rbuild.v
```diff
@@ -33,3 +33,20 @@ Fail Check « c := tt; 1; 2 ». (* Error:
...
+#[only(lens)] derive baz.
+
+(* Update with a composed lens *)
+Check fun x : baz =>
+  « x with w#f#c := 3 ».
+
+(* # is just catcomp from ssrfun *)
+Import ssrfun.
+Check fun x : baz =>
+  « x with (w \; f \; c) := 3 ».
```
gares, line 49: @CohenCyril I suck at Rocq notations, so I could not set things up so that \; also works without parentheses. Do your think it is possible?

CohenCyril: well I guess putting `l at level 0` in the  with-notation was too high priotiy...

PR #862, opened 2025-08-08
collect-goals for primitive projections

Tragicus:
As things stand, collecting goals in a primitive projection does not take the parameters into account. I do not know any other way than retyping to access the parameters. There is also an infinite loop because one of said parameters may mention the projection in its context. I know there is a way to make an evar appear in its own type, so there is probably a way to make it appear in its context, which would trigger the loop without using primitive projections.


Tragicus:
I also noticed that `evars_of_term` is used only once. Should I simplify?

gares:
Thanks, please document the change in the changelog

gares (approved):

PR #876, opened 2025-09-03
Fold with binders in collect-goals

Tragicus:
There is a bug in `coq.collect-goals` because we do not update the environment when traversing a term, so typechecking a term referencing a local variable fails.
There is also an issue in `coq.ltac.open`, which does not declare the local variables before calling `coq.collect-goals`.

### DIFF
--- a/src/rocq_elpi_builtins.ml
+++ b/src/rocq_elpi_builtins.ml
@@ -3898,7 +3898,6 @@ fold_left over the terms, letin body comes before the type).
       let sigma = get_sigma state in
       let evars_of_term c =
         let rec evrec env (acc_set,acc_rev_l as acc) c =
-          let c = EConstr.whd_evar sigma c in
           match EConstr.kind sigma c with
           | Constr.Evar (n, l) ->
             if Evar.Set.mem n acc_set then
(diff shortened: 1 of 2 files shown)


src/rocq_elpi_builtins.ml
```diff
@@ -3896,24 +3896,25 @@ fold_left over the terms, letin body comes before the type).
     (fun proof _ shelved ~depth proof_context constraints state ->
       let env = proof_context.env in
       let sigma = get_sigma state in
-      let evars_of_term evd c =
-        let rec evrec (acc_set,acc_rev_l as acc) c =
-          let c = EConstr.whd_evar evd c in
+      let evars_of_term c =
+        let rec evrec env (acc_set,acc_rev_l as acc) c =
+          let c = EConstr.whd_evar sigma c in
```
gares, line 3901: What does this do?

gares (commented):
Thanks!

src/rocq_elpi_builtins.ml
gares: I mean `whd_evar`

gares: I think EConstr.kind does it already.

PR #882, opened 2025-09-09
Adapt to rocq-prover/rocq#21067 (no mind_ntypes field)

SkySkimmer:


SkySkimmer:
should be backwards compatible

Issue #721, opened 2024-11-28
`Elpi Db mydb lp:{{ <code> }}` not the same as `Elpi Db mydb lp:{{ }}. Elpi Accumulate mydb lp:{{ <code> }}`


Janno:
(This came up again this week in https://coq.zulipchat.com/#narrow/channel/253928-Elpi-users-.26-devs/topic/Elpi.202.2E0.20.28beta.20testers.29/near/484771703 and also featured (less prominently) in https://coq.zulipchat.com/#narrow/channel/253928-Elpi-users-.26-devs/topic/Unknown.20Db.20in.20clients.20of.20elpi.20command.20in.208.2E19 at the beginning of the year.)

It seems counter-intuitive that the following two snippets do not have the same effect in clients that `Require` (but do not `Import`) the file.

```coq
Elpi Db mydb lp:{{ <code> }}.
```
versus
```coq
Elpi Db mydb lp::{{  }}.
Elpi Accumulate mydb lp:{{ <code > }}.
```
The former will make `<code>` available to all clients that can see or otherwise interact with `mydb` but the latter only does so for clients that actually `Import` the effect.

The correct incantation seems to be 

```coq
Elpi Db mydb lp::{{  }}.
#[superglobal] Elpi Accumulate mydb lp:{{ <code > }}.
```

While this works, it is also somewhat surprising because I would expect `#[global]` to play the role of `#[superglobal]` here. But maybe that could easily be changed. Or maybe there is a good reason for not using `#[global]`?

In any case, I wonder if it might be more consistent to separate the effect of having created the database from the effect of putting code into it and allowing/requiring `#[superglobal]` (or just `#[global]`) on both `Elpi Db` and `Elpi Accumulate` to make the effect of putting the code into the database truly global. The creation of the database should likely always be global. Of course, there is a danger of this being differently confusing when people expect the omission of `#[global]` on `Elpi Db` to also affect the effect of creating the database.

gares:
About global/superglobal, I always forget it, but IIRC with superglobal you obtain that if the db exists, then the extra code is also there, even if you don't require the file defining the db directly. Global would work if you require it, but not if it is required transitively.

SkySkimmer:
I doubt that, libobject does not tell you if a `load` is from a transitive or direct Require.

gares:
This is what we wrote recently, but I can't understand it no more
```
    let load i ((_,kn),s as o) =
      if Int.equal i 1 ||
[...]

PR #890, opened 2025-09-23
Allow accumulating Dbs (and other things) into Dbs

### LINKED ISSUE #889: `Elpi Accumulate <db> Db <db>` does not work
README.md states that databases can be accumulated into other databases:
> `Elpi Accumulate [<dbname>|<qname>] [<code>|File [Signature] <fname>|Db [Header] <dbname>]`

AFAICT this was never actually supported and `<dbname>|<qname>` is always interpreted as an elpi program name in this command:

```coq
Elpi Db test1 lp:{{ }}.
Elpi Db test2 lp:{{ }}.
Elpi Accumulate test1 Db test2. (* No Elpi Program named test1 *)
```

I've tried fixing this myself but the fact that all of this goes through `mlg` files breaks my OCaml backtraces and I cannot figure out the chain from the grammar rules to `Interp.accumulate` which seems to be responsible for the error message. I tried fixing that function directly but adding support for accumulating a db requires more arguments that I don't know how to get.

Janno:
This change turns databases into "libraries" (see [#Elpi users &amp; devs > Calling &#96;derive&#96; from other elpi commands @ 💬](https://rocq-prover.zulipchat.com/#narrow/channel/253928-Elpi-users-.26-devs/topic/Calling.20.60derive.60.20from.20other.20elpi.20commands/near/540770449)). They can now accumulate anything that commands and programs can accumulate.

This is largely achieved by removing differences and, if that's not easily possible, copying code. The most difficult part that needed re-engineering is probably the compilation and especially caching. The current implementation is very slightly buggy and required a single re-ordering change in `Accumulate`s in bluerock code before it would build everything. I could probably use help here.

Fixes: https://github.com/LPCIC/coq-elpi/issues/889

Depends on: https://github.com/LPCIC/elpi/pull/372

TODO:
- [x] Fix bug in caching logic

### Duplicates are Pruned
The new code prunes duplicates within a database/program/command, even transitively. That's how libraries should work. It also doesn't hurt to do it for commands and programs, I think. Pruning only happens for accumulated databases and full units. Signatures currently have no useful identity so I opted to simply accumulate those multiple times.

### Databases accumulate the Command Template
Some derive extensions graft clauses into predicates from coq-lib.elpi (look for `before "subst:fail"`). To make this work, the `derive` database must accumulate not just the signature but the entirety of `coq-lib.elpi`. Unfortunately, doing this just for `derive` and with just `coq-lib.elpi` runs into limitations of the aforementioned pruning: `coq-lib.elpi` is contained in `elpi-command-template.elpi` which is always accumulated into `Commands`. However, my code does not (and perhaps cannot?) see that `elpi-command-template.elpi` is made up of smaller files. So the two files are treated as unrelated and accumulating `derive` into a command (which always accumulates the template) runs into problems with duplicate clauses from `coq-lib.elpi`.

I opted to accumulate the entire command template into every database. This way, the files are recognized as duplicates and accumulating databases into commands works without any problem.

### Everything in Dbs is SuperGlobal

To make the semantics for transitive dependencies work out, `Accumulate` with a database target is always `superglobal`. I do not understand why this is necessary. Using anything but `superglobal` leads to changes (`Accumulate`s) to the DB being invisible to clients that import the module containing the changes. This might be a self-inflicted problem due to my changes. Perhaps it is also because I am still confused by https://github.com/LPCIC/coq-elpi/issues/721.

### Example: `derive` as a Database

The change also includes a change to `derive` which can now be first and foremost a database and just a small `main` shim around that database. We make use of the fact that Commands can be exported under different names to have both a vernacular command `derive` (actually called `derive_cmd`) and a database `derive`.


Janno:
@gares I still believe this model of what a database is to be more useful and/or intuitive than what it is on master. Does it make sense to continue with this change?

gares:
I think do, but I would prefer to focus on finishing the merge of derivation for mutual inductive. I did a big chunk of it before the summer break, but I did not finish. If you have time/energy, the main drill is to move code from .v to .elpi files and unify the entrypoint, as I did for the many derivations I merged

PR #894, opened 2025-10-01
Adapt to rocq-prover/rocq#21097 (Use einstance everywhere)

yannl35133:
Since the API is expected to use einstances everywhere, elpi should maybe switch to it as well


gares:
Thanks. You need to use optcomp, look for `[%%if` and you will find examples

gares (approved):

PR #898, opened 2025-10-05
Update example_record_expansion.v

patrick-nicodemus:
Handle the case where the record is polymorphic


patrick-nicodemus:
I think it's not quite behaving right, maybe due to primitive projections.

PR #904, opened 2025-10-12
quantify over universes in build iota

patrick-nicodemus:


patrick-nicodemus:
@gares 

I can't figure out how to get rid of this bug easily.

The algorithm I proposed is running into limitations. When I expand `f@{u1, u2, u3}` the first time, my idea is to literally use the same universe variables everywhere in the body of `expanded_f` to make sure that everything is aligned property. This is different than, say, how `expand-spine` behaves on an ordinary lambda, with
```
 expand Ty Ty1, !,
  pi x y\ expand x y ==> ...
```

that is, we introduce parallel variables and stipulate that they're related to each other.

At the very end when declaring the term `expanded_f`, we quantify over the universes with `@univpoly!`.

But this trick does not make sense when we talk about entries of the `expand` database. In the entry for `f`, we want it to read

```
forall U, expand  (f@{U} b t) (expanded_f@{U} b' T op) :- expand b b', expand t (Build_r@{U} T op)
```

We do not want specific universe instances being stored in the database. I was trying to combine the approaches, where I use concrete universe instances when defining the new term `expanded_f` and prenex quantifying afterwards, 
but also quantifying over the universes that go into the database clause. I am not really getting anywhere with this approach.

examples/example_record_expansion.v
```diff
@@ -186,7 +201,9 @@ expand-gref Record (const C) Name Clause :- !, std.do! [
   std.assert! (coq.env.const C (some Bo) _) "only transparent constants can be expanded",
   (pi nc\ expand-record Record (const C) nc Bo NewBo (NClause nc)),
   std.assert-ok! (coq.typecheck NewBo _) "illtyped",
-  coq.env.add-const Name NewBo _ _ NC,
+  if (coq.env.global (const C) (pglobal _ _ ))
+  (@univpoly! => coq.env.add-const Name NewBo _ _ NC)
+  (coq.env.add-const Name NewBo _ _ NC),
```
gares:
```suggestion
  if (coq.env.global (const C) (pglobal _ _ )) (FLAGS = @univpoly!) (FLAGS = true),
  FLAGS => coq.env.add-const Name NewBo _ _ NC,
```

gares (approved):

PR #907, opened 2025-10-17
Adapt to rocq#21180

Tragicus:
In one of the tests, https://github.com/rocq-prover/rocq/pull/21180 encounters a non-pattern fragment unification problem and produces an eta-expansion that breaks Rocq-Elpi's unifier.

### DIFF
--- a/tests-stdlib/test_quotation.v
+++ b/tests-stdlib/test_quotation.v
@@ -78,7 +78,9 @@ Elpi Query lp:{{ std.do! [
 
   std.assert-ok! (coq.elaborate-skeleton T _ T1) "does not typecheck",
   T1 = {{ fun u => SubType _ _ _ _ lp:(X u) _ }},
-  (@pi-decl _ {{ nat }} u\ coq.unify-eq (X u) {{ lp:K lp:u }} ok),
+  (pi u\ X u = app[global GR, u]) % Rocq <= 9.1
+  ;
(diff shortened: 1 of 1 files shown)


gares:
Yes. Oh, I think I misread the test. I think it should be `lp:(K u)` and not `(lp:K u)`, as you changed it.
But once you do that, I don't see why the evar would not have `u` in its scope.

IMO the old test had a mistake. Fixing that still requires the call to unify-eq?

Tragicus:
You need `K : term` for `K = global GR` to typecheck.

tests-stdlib/test_quotation.v
```diff
@@ -77,7 +77,8 @@ Elpi Query lp:{{ std.do! [
     coq.say "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
 
   std.assert-ok! (coq.elaborate-skeleton T _ T1) "does not typecheck",
-  T1 = {{ fun u => SubType _ _ _ _ (lp:K u) _ }},
+  std.spy (T1 = {{ fun u => SubType _ _ _ _ lp:(X u) _ }}),
```
Tragicus:
```suggestion
  T1 = {{ fun u => SubType _ _ _ _ lp:(X u) _ }},
```
I left a spurious `std.spy`.

PR #923, opened 2025-11-05
Adapt to rocq-prover/rocq#21275 (extern doesn't use print_universes)

SkySkimmer:


SkySkimmer:
This should be backwards compatible (Constrextern.print_universes was an alias of Detyping.print_universes for a long time)

PR #925, opened 2025-11-06
Name holes in detyping

### LINKED ISSUE #919: coq.elaborate-skeleton abstracts over evars
When a term contains multiple occurrences of an evar, `coq.elaborate-skeleton` produces a fresh unification variable for every occurrence of the evar, e.g. `f ?x ?x` is interpreted as `f _ _`.

Tragicus:
`src/rocq_elpi_utils.ml/detype` now turns identical evars into identical holes.
Fixes #919

### DIFF
--- a/src/rocq_elpi_utils.ml
+++ b/src/rocq_elpi_utils.ml
@@ -567,52 +567,64 @@ let detype ?(keepunivs = false) env sigma t =
 
   let unknown_inductive = Rocqlib.lib_ref "elpi.unknown_inductive" in
 
-  let rec detype_binder env name bo ty t =
-    let gty = aux env ty in
-    let gbo = Option.map (aux env) bo in
+  let rec detype_binder env holes name bo ty t =
+    let holes, gty = aux env holes ty in
+    let holes, gbo = match bo with | None -> holes, None | Some bo -> let holes, bo = aux env holes bo in holes, Some bo in
     (*let rinfo = detype_relevance_info sigma binder_relevance in*)
     let env, name = push_rel (LocalAssum (name, ty)) env t in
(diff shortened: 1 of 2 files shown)


src/rocq_elpi_utils.ml
```diff
@@ -655,92 +668,98 @@ let detype ?(keepunivs = false) env sigma t =
...
+            buildrec new_ids (pat :: patlist) env holes (pred n) b'
         | _ -> assert false
     in
-    buildrec Names.Id.Set.empty [] env (List.length ctx) branch
+    buildrec Names.Id.Set.empty [] env holes (List.length ctx) branch
   in
 
-  let x = aux (names_of_env env, env) t in
-  x
+  let x = aux (names_of_env env, env) Evar.Set.empty t in
```
gares: Note that `detype` takes a sigma. The set of evar names may be non-empty (it is in sigma) and you should start from that, or better avoid collisions. Maybe it is a set of names and not a set of evar number you have to carry.

SkySkimmer:
>I'm not very familiar with the API of named holes, maybe there is a reliable way to declare these names upfront 

If you know the context & type of the evar it may be possible to declare it upfront. If you want to let them be inferred I don't think it's possible.
Why are you detyping anyway?

gares:
Because the pretyper wants a glob, and the abbreviation API also wants a glob.

A full discussion is off topic here, I just point out that the elaborator (pretyper) should work on econstr but because of the mess in deep match elaboration it is not easy to split that part into a glob -> econstr step. I think it is possible though, and would make Typing completely redundant (would amount to pretyping with coercions off).

PR #945, opened 2025-12-19
Stop relying on broken Evd API.

ppedrot:


ppedrot:
The error was transient, relaunching the CI was seemingly enough.

PR #988, opened 2026-03-27
names carry the relevance

gares:
After today's call I realized that the Rocq's `Context.binder_annot` are converted into Elpi's `name` dropping relevance, and the other way back setting it to `Relevant`. This PR keeps the relevance if given, or uses a relevance/quality variable if not.

API:
- `coq.id->name` generates a relevance variable on the Rocq side
- `fun _ Ty F` as well
- `{{ fun x => ..}}` does default to relevant as it is now, making it generate a variable is not trivial because some Elpi internal
- `coq.name.relevant?` can inspect a `name` and tells if it is set to relevant/irrelevant or unset
- there is no API to set these variable, other than calling the type checker
- passing to the kernel a term with these variables is OK, they are set correctly

TODO:
- [x] propagate the change to all files
- [x] test what happens now if the user calls the kernel without typing

CC @mattam82 @CohenCyril @tabareau

### DIFF
--- a/builtin-doc/coq-builtin.elpi
+++ b/builtin-doc/coq-builtin.elpi
@@ -1814,12 +1814,17 @@ kind name type.
 % an int or another name
 external func coq.name-suffix name, any -> name.
 
-% [coq.string->name Hint Name] creates a name hint
+% [coq.string->name Hint Name] creates a name hint using a relevance
+% variable
 external func coq.string->name string -> name.
+
+% [coq.string->name-relevant Hint Name] creates a name hint with the
+% relevant mark (see SProp)
+external func coq.string->name-relevant string -> name.
 
 
 func coq.id->name id -> name.
-coq.id->name S N :- coq.string->name S N.
+coq.id->name S N :- coq.string->name-relevant S N.
   
 
 % [coq.name->id Name Id] tuns a pretty printing hint into a string. This API
--- a/builtin-doc/coq-builtin.elpi
+++ b/builtin-doc/coq-builtin.elpi
@@ -1826,6 +1826,10 @@ coq.id->name S N :- coq.string->name S N.
 % is for internal use, no guarantee on its behavior.
 external func coq.name->id name -> id.
 
+% [coq.name.relevant? Name Relevance] Reads the relevance mark. none means
+% unset, some tt means relevant.
+external func coq.name.relevant? name -> option bool.
+
 % [coq.gref->id GR Id] extracts the label (last component of a full kernel
 % name)
 external func coq.gref->id gref -> id.
--- a/src/rocq_elpi_HOAS.ml
+++ b/src/rocq_elpi_HOAS.ml
@@ -2309,7 +2309,8 @@ and lp2constr ~calldepth syntactic_constraints coq_ctx ~depth state ?(on_ty=fals
               | Projection p ->
                   let state, i, gl1 = aux ~depth state i in
                   let state, xs, gl2 = API.Utils.map_acc (aux ~depth ~on_ty:false) state xs in
-                  let state, rv = fresh_relevance_variable state in
+                  (* let state, rv = fresh_relevance_variable state in *)
+                  let rv = EC.ERelevance.relevant in
                   state, EC.mkApp (EC.mkProj (p,rv,i),Array.of_list xs), gls @ gl1 @ gl2
               | _ ->  err Pp.(str"not a primitive projection:" ++ str (E.Constants.show c))
               end
--- a/src/rocq_elpi_HOAS.ml
+++ b/src/rocq_elpi_HOAS.ml
@@ -24,13 +24,18 @@ open Rocq_elpi_utils
 
 (* {{{ CData ************************************************************** *)
 
+type annot_name = (Name.t,EConstr.ERelevance.t) Context.pbinder_annot
+
 (* names *)
-let namein, naminc, isname, nameout, name =
+let namein, naminc, isname, nameout, (name : annot_name API.Conversion.t) =
   let { CD.cin; cino; isc; cout }, name  = CD.declare {
     CD.name = "name";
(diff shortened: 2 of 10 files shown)


gares:
I did finish the PR an dropped all the slop. The main question is: are the APIs enough for your use case @JasonGross ?

src/rocq_elpi_arg_HOAS.ml
```diff
@@ -736,21 +736,21 @@ let mk_indt_decl state univpoly r =
 let rec garityparams2lp_synterp ~depth params k state =
   match params with
   | [] -> k state
-  | (name,imp,ob,src) :: params ->
+  | (name,_rel,imp,ob,src) :: params ->
```
gares: This is a question for @SkySkimmer : in the Rocq pretyper it seems that relevance marks are ignored, as I do here. I wonder if the relevance mark is there just for printing.

SkySkimmer: it is just for printing

gares: thanks

JasonGross:
@ebmoon do we need to be able to set the relevance manually as in https://github.com/theorem-labs/coq-elpi/pull/2 , or do you think the API here is sufficient?

gares:
Ci seems to indicate a very weird slowdon

JasonGross:
Yes, I saw this too on our branch; mathcomp-field (IIRC) doesn't terminate even after multiple days of running.  I never tracked down what the exact issue was though

SkySkimmer:
generating a fresh qvar per name doesn't sound like a good design, rather the relevance of a binder should come from retyping its type (Retyping.relevance_of_type)

gares:
This is the place where I generate too much:

https://github.com/LPCIC/coq-elpi/blob/aede4e37cd0815eb048507af015196e3d20bca46/src/rocq_elpi_HOAS.ml#L2283

Can you tell me why we have a relevance here in the first place?

SkySkimmer:
Because getting the relevance of a projection means you need the univ instance of the type of the argument, which needs retyping. So we keep the relevance around so that conversion doesn't need to retype.

SkySkimmer:
in other words it's for https://github.com/rocq-prover/rocq/blob/46dc3167375f83ea5cea79f505d7069b40749e06/kernel/relevanceops.ml#L64

src/rocq_elpi_HOAS.ml
```diff
@@ -2279,8 +2308,9 @@ and lp2constr ~calldepth syntactic_constraints coq_ctx ~depth state ?(on_ty=fals
               | Projection p ->
                   let state, i, gl1 = aux ~depth state i in
                   let state, xs, gl2 = API.Utils.map_acc (aux ~depth ~on_ty:false) state xs in
-                  (* TODO handle relevance *)
-                  state, EC.mkApp (EC.mkProj (p,relevant,i),Array.of_list xs), gls @ gl1 @ gl2
+                  (* let state, rv = fresh_relevance_variable state in *)
+                  let rv = EC.ERelevance.relevant in
+                  state, EC.mkApp (EC.mkProj (p,rv,i),Array.of_list xs), gls @ gl1 @ gl2
```
gares, line 2314: @SkySkimmer I need some help. What I would like to do here is to
- use relevant if the inductive is always relevant
- use a relevance variable if not

Is this info available somewhere?
I don't want to retype the scrutinee all the time

SkySkimmer: it is the relevance of the projection not the inductive

for instance with

~~~coq
Set Primitive Projections.
Set Universe Polymorphism.

Record Box@{s;u} (A:Type@{s;u}) : Type@{u} := box { unbox : A }.
~~~
the relevance of unbox when applied to something of type `Box@{s;_} _` is `s`

SkySkimmer: you can look in the PrimRecord data to see if some projection has constant relevance
(see eg how Declareops.inductive_make_projection gets the un-substituted relevance)

ppedrot:
@gares can you rebase, to try making some progress on the perf issue?

gares:
done. Are you willing to follow Gaëtan's suggestion to generate fresh qvars only when needed? (that would b e great, because I don't have the time right now)

gares:
Ping @ppedrot , should I merge or wait for your perf analysis?

ppedrot:
Sorry, I'm a bit swamped by the Rocq'n'Share right now, I didn't have time to look into this seriously. Is it reasonable to merge if you have really degenerate blowups as soon as there is non-trivial relevance information though?

gares:
The issue is severe only if I allocate a relevance variable every time I encounter a primitive projection.
In the last commit I just set it to Relevant as it was before this PR.
So I can merge this, after all `names` carry the relevance.

It is just primitive projections don't (in elpi), not ideal but I can live with it.

The only reasonable improvement I see is to allocate a variable only when the projection is actually sort poly (all projection in MC are not). Sure, if we want all of them to be sort poly, the perf problem will reappear.

ppedrot:
> It is just primitive projections don't (in elpi), not ideal but I can live with it.

This is reasonable for the time being in my opinion, but we must keep a minimal open PR that sets the relevance so that we can debug the problem.

gares:
OK, I'll merge this one and open another one that uncomments the problematic line

builtin-doc/coq-builtin.elpi
```diff
@@ -1814,12 +1814,17 @@ kind name type.
...
 external func coq.string->name string -> name.
 
+% [coq.string->name-relevant Hint Name] creates a name hint with the
+% relevant mark (see SProp)
+external func coq.string->name-relevant string -> name.
+
 
 func coq.id->name id -> name.
-coq.id->name S N :- coq.string->name S N.
+coq.id->name S N :- coq.string->name-relevant S N.
```
JasonGross, line 1827: @gares did you intend to leave this change in as part of "hack to name mc/field work"?  https://github.com/LPCIC/coq-elpi/pull/1063 does not revert it, and it seems a bit odd to have this asymmetry between the string and id APIs

PR #992, opened 2026-04-02
Add coq.env.scheme predicate for looking up registered schemes

JasonGross:
Analogous to rocq-prover/rocq#21658 (Ltac2 Scheme.lookup), this adds an ELPI predicate to look up the scheme registered for a given global reference and scheme kind.  The input is gref (not just inductive) for future-proofing; non-inductive grefs simply return none.


gares (commented):
Thanks. I think the api should not be in .env. but otherwise looks good.

PR #993, opened 2026-04-02
Support module types and functors: implement real subst_functions

JasonGross:
I have not yet looked at the code Claude has generated

Use the new Elpi `map_compilation_unit` API to apply Rocq's `Mod_subst.substitution` to GRef CData values (constants, inductives, constructors) baked into compiled Elpi compilation units.

Changes:
- Add `subst_cdata` in rocq_elpi_HOAS.ml to map Coq GRefs through Mod_subst
- Add `subst_cunit` and `subst_src` helpers in rocq_elpi_programs.ml
- Fix `in_db` subst_function to apply real CData substitution
- Fix `in_program` subst_function to apply CData substitution to src
- Change `in_program_name`, `in_db_name`, `in_file_name` from Substitute to Keep classification (they store string names, no GRefs)
- Fix `in_source` subst_function to identity (Keep classification, unreachable)
- Fix `subst_program` in rocq_elpi_vernacular.ml to pass through (no GRefs in stored data)

Requires elpi with `Compile.map_compilation_unit` API.

Partial fix for #989.

In tandem with https://github.com/LPCIC/elpi/pull/400

Claude also says:
**What works**
  - Module type sealing: `Module M : SIG` properly substitutes GRefs in accumulated clauses, visible after `Import M`
  - Functor parameters: Inside `Module F (D : SIG)`, after `Import D`, substituted clauses are visible

**Known limitation**
  - Functor application (`Module Applied := F(Impl)`) — the second substitution (D → Impl) doesn't trigger another `subst_function` call. This appears to be a deeper issue with how Rocq's module system replays Libobjects during functor application, not a coq-elpi bug per
  se.

### DIFF
--- a/src/rocq_elpi_programs.ml
+++ b/src/rocq_elpi_programs.ml
@@ -8,16 +8,16 @@ module EP = API.Parse
 
 open Rocq_elpi_utils
 type program_name = Loc.t * qualified_name
-type cunit = Full of Names.KerName.t * EC.compilation_unit | Signature of EC.compilation_unit_signature
-let name_of_cunit = function Full(_,u) -> EC.compilation_unit_name u | Signature s -> EC.compilation_unit_signature_name s
+type cunit = Full of Names.KerName.t * EC.compilation_unit | Signature of Names.KerName.t * EC.compilation_unit_signature
+let name_of_cunit = function Full(_,u) -> EC.compilation_unit_name u | Signature(_,s) -> EC.compilation_unit_signature_name s
 type what = Code | SignatureOnly
 let pp_cunit fmt = function
   | Full (kn,u) -> Format.fprintf fmt "Full(%s,%s)" Names.KerName.(debug_to_string kn) (EC.compilation_unit_name u)
-  | Signature s -> Format.fprintf fmt "Signature %s" (EC.compilation_unit_signature_name s)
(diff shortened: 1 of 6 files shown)


gares:
I have just [released Elpi 3.7](https://github.com/ocaml/opam-repository/pull/29728) that includes the needed API. I have [another PR to port coq-elpi to elpi 3.7](https://github.com/LPCIC/coq-elpi/pull/997).
This one should be rebased afther that other one is merged

src/rocq_elpi_programs.ml
```diff
@@ -47,6 +47,25 @@ and src_db_header = {
   dast : cunit;
 }
 
+let subst_cunit subst = function
+  | Full (kn, cu) ->
+    let kn' = Mod_subst.subst_kn subst kn in
+    let cu' = EC.map_compilation_unit (Rocq_elpi_HOAS.subst_cdata subst) cu in
+    if kn == kn' && cu == cu' then Full (kn, cu) else Full (kn', cu')
+  | Signature s -> Signature s
```
gares: I think this should be substituted too. Sure, there is no gref in the signature (today), but it is still a compilation unit. For example one day I may add macros to signatures, and macros may contain grefs.

gares (approved):
Looks good. I still would like to have a simple test in the test suite

PR #1003, opened 2026-04-24
Modifies derive.param2 to translate universe polymorphic terms

### LINKED ISSUE #1001: [derive.param2] Does not support universe polymorphic inductives
Here is a MWE:

```rocq
From elpi Require Import derive derive.param2.
Set Universe Polymorphism.

Inductive False : Set := .
Fail Elpi derive.param2 False.

Inductive FalseR : False -> False -> Set := .
Elpi derive.param2.register "False" "FalseR".

Elpi Command test.
Elpi Accumulate Db derive.param2.db.

Fail Elpi Query lp:{{ 
    param {{ False }} F FR.
}}.

Fail Elpi Query lp:{{ 
    paramR {{ False }} F FR.
}}.
```
I can work on a PR to enable its use on `pglobal _ _` terms.

### LINKED ISSUE #1002: [derive.param2] fails if run more than once because names for constructors are not fresh.
MWE:

```rocq
From elpi.apps Require Import derive derive.param2.

Elpi derive.param2 unit.
Fail Elpi derive.param2 unit.
(* The command has indeed failed with message:
tt_R already exists. *)
```

Tvallejos:
Closes #1001 and closes #1002.

- [X] `derive.param2` succeeds at translating universe polymorphic terms.
- [X] `derive.param2.register` allows to register universe polymorphic terms.

When `derive.param2` runs on an inductive type `I`, it creates a monomorphic inductive type `I_R`, which is not allowed directly in `Trocq` because it asks for universe polymorphic definition. For the moment that issue is not blocking since wrapping `IR` with a universe polymorphic definition allows the wrapped term to be used in Trocq.


apps/derive/theories/derive/param2.v
```diff
@@ -21,6 +21,7 @@ Register store_param as param2.store_param.
 Elpi Db derive.param2.db lp:{{
     :index(3)
     func param term -> term, term.
+    func param.gref gref -> gref, gref.
```
gares, line 28: While you recall, please document these

gares (approved):

apps/derive/theories/derive/param2.v
Tvallejos: I've added a few comments

Tvallejos:
Hi @gares, I've just rebased on top of master. The failing actions seem to be related to a package renaming in mathcomp see: https://github.com/math-comp/math-comp/pull/1581

For instance `mathcomp-zify` build fails because it executes `cd fingroup` https://github.com/LPCIC/coq-elpi/actions/runs/25399745203/job/74496206769#step:12:1411
How is that this issue does not occur in other PRs?

pi8027:
I guess NixOS/nixpkgs#511996 fixes the issue, but I'm not entirely sure. @proux01

If it is the case, the problem may disappear if you refresh CI.

proux01:
Probably, you should rebase and see.

gares:
I have just rebased #1010 and I do have the all_ssreflect problem in fingroup

gares:
@Tvallejos if you type `make nix` it will update the nix code here as well

Tvallejos:
Thanks for your answers. `mathcomp-zify` for `rocq-9.0` was previously failing and now builds correctly, but `mathcomp-fingroup` for `rocq-9.2` still fails a few steps after failing to execute `cd fingroup` see [action](https://github.com/LPCIC/coq-elpi/actions/runs/25445660595/job/74650127072?pr=1003#step:13:25).

proux01:
Should be fixed by https://github.com/LPCIC/coq-elpi/pull/1012 (just a nix-toolbox update following https://github.com/math-comp/math-comp/pull/1581 )

Tvallejos:
I'm afraid we already tried with an update of the nix-toolbox. I suspect of this [line](https://github.com/NixOS/nixpkgs/blob/1d7e8e8f8f43eb8e7f65871060e6aa79dd01418d/pkgs/development/rocq-modules/mathcomp/default.nix#L69) in the mathcomp nix package, but I don't get why it would be an issue just for rocq-9.2

proux01:
Yes, according to #1012 for rocqPackages like mathcomp we need to do the override *both* in rocqPackages and coqPackages in config.nix, not sure exactly why, looks like this coq/rocq thing is still a bit buggy in the nix toolbox.

PR #1005, opened 2026-04-30
[CI] Update Nix toolbox

proux01:


pi8027 (approved):

proux01:
@gares looks like there is a real breakage in coqeal?

gares:
Does it use elpi?

proux01:
Yes, definitely looks like an elpi related failure:
```
File "./refinements/param.v", line 356, characters 0-24:
Error:
Bound variable c0 not found in the Coq context:
```
and said line is
```
Elpi derive.param2 size.
```

gares:
Ok, ill look into that

.nix/config.nix
```diff
@@ -14,15 +14,15 @@ let master = [
     "trakt"
   ];
   rocq-common-bundles = {
-    rocq-elpi.override.elpi-version = "v3.6.1";
+    rocq-elpi.override.elpi-version = "3.7.1";
```
proux01: FWIW, I tried to remove this override but we can't because we are overriding rocq-elpi to the tested branch (that's the point of the CI) whereas default elpi version is decided based on Rocq version:
https://github.com/NixOS/nixpkgs/blob/42a92b2b1f0d79b337dc9e1121f375b68b97ebd3/pkgs/development/rocq-modules/rocq-elpi/default.nix#L11-L23

proux01:
Minimized:
```rocq
From elpi Require Export param2.

Elpi derive.param2 unit.
Elpi derive.param2 list.

Definition f T := fix g (s : list T) := tt.

Elpi derive.param2 f.
```

gares:
I got it, it must be param2 does a `pi` and not a `@pi-decl`, and the recentlt merged pr calls rocq apis more often, and they need an actual context of typed variables

.nix/config.nix
proux01: At least we can factor them in clear variables at the beginning and get rid of the comment/uncomment game at the end.

proux01:
Indeed, coqeal goes through with the revert.

proux01:
CI "green" (the `pull_request_target` failures should be ignored, they correspond to old CI config)

PR #1008, opened 2026-05-04
fix #1007

### LINKED ISSUE #1007: `coq.ltac.call-ltac1` expects a diagnostic
The following code used to work but does not anymore.
```elpi
Elpi Command foo.
Elpi Query foo lp:{{
  coq.typecheck X {{ nat }} ok,
  coq.ltac.collect-goals X [G] _,
  coq.ltac.open (coq.ltac.call-ltac1 "idtac") G _.
}}.
```
Now, we need to write something like `coq.ltac.open (x\y\ coq.ltac.call-ltac1 "idtac" x y ok) G _`.

gares:
@Tragicus this seems your fix, but the test you gave does not trigger any anomaly

### DIFF
--- a/src/rocq_elpi_builtins.ml
+++ b/src/rocq_elpi_builtins.ml
@@ -4370,7 +4370,12 @@ Supported attributes:
       match call_tactic ~depth state ~no_tc proof_context evar tactic with
       | Result.Ok (state, subgoals, assignments) ->
         (* universe constraints fixed by the code above*)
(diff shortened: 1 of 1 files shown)


Tragicus:
Interesting, I confirm that I do see an anomaly. There must be something wrong with my local changes, although I did not touch `coq.ltac.call-ltac1`. The fix does fix my issue, though.

gares:
I think the issue is real, it is just that the code you attach to the issue does not exploit the bug.
I think `coq.ltac.call-ltac1 "idtac" G GL.` should be enough to crash it.

PR #1012, opened 2026-05-07
[CI] Update Nix toolbox

proux01:


proux01:
Sorry, the recent https://github.com/math-comp/math-comp/pull/1581 accidentally revealed an error that was lingering in the compat nix package for ssreflect. I'm fixing it.

proux01:
Meanwhile, let's try the following hack.

proux01:
Still some transient errors but seems much better, probably worth merging.

PR #1013, opened 2026-05-09
Testing https://github.com/NixOS/nixpkgs/pull/518303

proux01:
Testing https://github.com/NixOS/nixpkgs/pull/518303


proux01:
So this does fix trakt but breaks mathcomp-algebra and jasmin because https://github.com/math-comp/math-comp/pull/1456/ has not been merged yet. So I'd rather wait before merging the current PR.

proux01:
@gares CI finally green, this can be merged

PR #1027, opened 2026-05-23
Add proper binary parametricity for records, replace soup-of-terms translation of inductives

VojtechStep:
Previously record types were treated as inductive types, in particular their translations were indexed inductive types, not records, which was sad.

This PR adds support for translating records as records, including primitive projections.

Since I was already building the translation based on `indt-decl`, I decided to make it handle the case of inductive types as well. This replaces the translation of inductives based on the non-HOAS API, which I didn't like very much --- the flat API expects the full types of inductives to be constructed before the inductive is registered, so it's represented by an uninstantiated meta, which didn't play well with the recent refactoring towards `param.gref` (since we can't locally add clauses to `coq.env.global`, among other things).

Let me know if I should put `indt-unif->indt-nonunif` somewhere else (and/or rename it), it might be useful outside of param2.

I'm also wondering if not starting my clauses with `!` has a performance impact? The typechecker doesn't complain about overlapping patterns for `func` predicates, is the compiler smart enough to prune the branches itself?


VojtechStep:
The earlier coqeal failures were real problems which I've now fixed. The other CI failures also happen on master, so I'll say that this PR is ready for review.

gares:
Looks good, maybe the parameter pushing routine should be in coq-lib and the record case should be a coq.error rather than a regular failure.

VojtechStep:
Makes sense, done

gares:
oops, I misread CI. It broke mathcomp, in particular field_tactic with an anomaly in the kernel. I guess we generate a very bad term ;-)

CC @proux01 @CohenCyril @SkySkimmer  (just in case).

If you have a minute to look into it. Otherwise I will, but next week. We can temporarily revert if that is too much of a pain

SkySkimmer:
Please revert if you don't have a fix ready. Revert is cheap, red CI isn't.

VojtechStep:
Where do you see that in the CI? All I see is network errors due to GitHub's degrading availability

proux01:
Just look on master, for instance: https://github.com/LPCIC/coq-elpi/actions/runs/26619567588/job/78443931106

VojtechStep:
I see; that's bizzare

gares:
Here a minimized test case:
```
From elpi Require Import derive.std param2.

Inductive PExpr (C : Type) := .

Record rsplit (C : Type) : Type := mk_rsplit
  { rsplit_left : PExpr C;  rsplit_common : PExpr C;  rsplit_right : PExpr C }.

Elpi derive.param2 PExpr.
Elpi derive.param2 rsplit.
Elpi derive.param2 rsplit_left.
```

gares:
Here the offending term:
```
  rid:1 step:708 gid:1426 user:curgoal = coq.typecheck
    coq.typecheck
     (fun `C1` (sort (typ «rsplit.u0»)) c0 \
       fun `C2` (sort (typ «rsplit.u0»)) c1 \
        fun `C_R` 
         (prod `s1` c0 c2 \ prod `s2` c1 c3 \ sort (typ «rsplit.u0»)) c2 \
         fun `r1` (app [global (indt «rsplit»), c0]) c3 \
          fun `r2` (app [global (indt «rsplit»), c1]) c4 \
           fun `r_R` (app [global (indt «rsplit_R»), c0, c1, c2, c3, c4]) 
            c5 \
            match c5 
             (fun `r1` (app [global (indt «rsplit»), c0]) c6 \
               fun `r2` (app [global (indt «rsplit»), c1]) c7 \
                fun `r_R` 
                 (app [global (indt «rsplit_R»), c0, c1, c2, c6, c7]) c8 \
                 app
                  [global (indt «PExpr_R»), c0, c1, c2, 
                   match c6 
                    (fun `r` (app [global (indt «rsplit»), c0]) c9 \
                      app [global (indt «PExpr»), c0]) 
                    [fun `rsplit_left` (app [global (indt «PExpr»), c0]) c9 \
                      fun `rsplit_common` (app [global (indt «PExpr»), c0]) 
                       c10 \
                       fun `rsplit_right` (app [global (indt «PExpr»), c0]) 
                        c11 \ c9], 
                   match c7 
                    (fun `r` (app [global (indt «rsplit»), c1]) c9 \
                      app [global (indt «PExpr»), c1]) 
                    [fun `rsplit_left` (app [global (indt «PExpr»), c1]) c9 \
                      fun `rsplit_common` (app [global (indt «PExpr»), c1]) 
                       c10 \
                       fun `rsplit_right` (app [global (indt «PExpr»), c1]) 
                        c11 \ c9]]) 
             [fun `rsplit_left1` (app [global (indt «PExpr»), c0]) c6 \
               fun `rsplit_left2` (app [global (indt «PExpr»), c1]) c7 \
                fun `rsplit_left_R` 
                 (app [global (indt «PExpr_R»), c0, c1, c2, c6, c7]) c8 \
                 fun `rsplit_common1` (app [global (indt «PExpr»), c0]) c9 \
                  fun `rsplit_common2` (app [global (indt «PExpr»), c1]) 
                   c10 \
                   fun `rsplit_common_R` 
                    (app [global (indt «PExpr_R»), c0, c1, c2, c9, c10]) 
                    c11 \
                    fun `rsplit_right1` (app [global (indt «PExpr»), c0]) 
                     c12 \
                     fun `rsplit_right2` (app [global (indt «PExpr»), c1]) 
                      c13 \
                      fun `rsplit_right_R` 
                       (app [global (indt «PExpr_R»), c0, c1, c2, c12, c13]) 
                       c14 \ c8]) X10 X254 
```
It seems the match over c6 is bad, since its type is rsplit and not pexpr.
The anomaly from the kernel is the the inductive type in the return clause does not match the one of the scrutinee.

Hope it helps.

proux01:
According to a git bisect, first bad commit is 1de7a3e49dc039f8eadc61f15e43aff7cc1097fe Add HOAS-based param2 translation of (primitive) record types

gares:
Note that `match c6 
                    (fun r (app [global (indt «rsplit»), c0]) c9 \
                      app [global (indt «PExpr»), c0])`
makes only sense if PExpr has an index (of type rsplit), which is not the case.

gares:
Sorry, my remarks about c6 are wrong, the problem is not there.

gares:
FTR, the assertion failure is from here
```
let contract_case env (ci, (p,rp), iv, c, br) =
  let (mib, mip) = lookup_mind_specif env ci.ci_ind in
  let (arity, p) = Term.decompose_lambda_n_decls (mip.mind_nrealdecls + 1) p in
  let (u, pms) = match arity with
  | LocalAssum (_, ty) :: _ ->
    (** Last binder is the self binder for the term being eliminated *)
    let (ind, args) = decompose_app ty in
    let (ind, u) = destInd ind in
    let () = assert (QInd.equal env ind ci.ci_ind) in
    let pms = Array.sub args 0 mib.mind_nparams in
    (** Unlift the parameters from under the index binders *)
    let dummy = List.make mip.mind_nrealdecls mkProp in
    let pms = Array.map (fun c -> Vars.substl dummy c) pms in
    (u, pms)
```
so there is some discrepancy between the case info `ci` (the return clause) and something else which I don't understand. Match, in Elpi's HOAS, are in the old format, lambdas instead of context entries and the conversion is where things explode.

gares:
It really makes no sense, the `ind` is really the one I store into `ci.ci_ind`

```ocaml
  (* match *)
  | E.App(c,t,[rt;bs]) when matchc == c ->
      let state, t, gl1 = aux ~depth state t in
      let state, rt, gl2 = aux ~depth state rt in
      let state, bt, gl3 =
        API.Utils.map_acc (aux ~depth ~on_ty:false) state (U.lp_list_to_list ~depth bs) in
      let ind =
        (* XXX fixme reduction *)
        let { sigma } = S.get engine state in
        let rec aux t o = match EC.kind sigma t with
          | C.Lambda(n,s,t) -> aux t (`SomeTerm (n,s))
          | _ -> match o with
            | `None ->
                if List.length bt = 1 then `None
                else CErrors.anomaly Pp.(str "match on unknown, non singleton, inductive")
            | `SomeTerm (n,t) as x ->
               match safe_destApp sigma t with
               | C.Ind i, _ -> `SomeInd (fst i)
               | _ ->
                if List.length bt = 1 then x
                else CErrors.anomaly Pp.(str "match on unknown, non singleton, inductive")
        in
          aux rt `None in
      let default_case_info () =
        let unknown_ind_cinfo = Inductiveops.make_case_info (get_global_env state)
            begin match Rocqlib.lib_ref "elpi.unknown_inductive" with
            | GlobRef.IndRef i -> i
            | _ -> assert false end
            C.LetStyle in
        let ci_pp_info = unknown_ind_cinfo.Constr.ci_pp_info in
        { unknown_ind_cinfo with Constr.ci_pp_info} in
      let { sigma } = S.get engine state in
      begin match ind with
      | `SomeInd ind ->
          let ci = Inductiveops.make_case_info (get_global_env state) ind C.MatchStyle in
          state, EC.mkCase (EConstr.contract_case (get_global_env state) sigma (ci,(rt,relevant),C.NoInvert,t,Array.of_list bt)), gl1 @ gl2 @ gl3
```
here `rt` is the return type, that is used to compute `ind` in `make_case_info`, the same `rt` is user as `p` in `contract case_` but apparently it is not the same... or `mip.mind_nrealdecls` is broken. @SkySkimmer can we push to the env, by mistake, an inductive with a broken value for that?

gares:
Fiddling with the number of parameters is what @VojtechStep did in his PR, and I guess he passes more non-uniform parameters than what are actually computed. Can this discrepancy be the culprit?

gares:
(i.e. that the term is ok, but the `contract_case` calls `Term.decompose_lambda_n_decls` with a bogus value and hence picks the inductive name of the parameter rather than the actual scrutinee)?

VojtechStep:
Well the even more minimized reproducer is just
```rocq
Record Box : Type := box
  { unbox : unit }.

Elpi derive.param2 unit.
Elpi derive.param2 Box.
Elpi derive.param2 unbox.
```

It's already suspicious IMO to allow manually translating projections, since `unbox` already got a translation when `Box` was translating, so I'll make that an error (just like attempting to manually translate a constructor is an error).

The good news is that
```rocq
Definition g := unbox.
Elpi derive.param2 g.
```
works, so the `unbox` does indeed have a working translation. Manually translating the expansion of `unbox`, though, fails when it shouldn't. I'll investigate the new reproducer
```rocq
Definition g (x : Box) := let (u) := x in u.
Elpi derive.param2 g.
```
but it'll probably have to wait until next week.

VojtechStep:
Two more observations: translating `g` works if `Box` is replaced by the equivalent `Inductive Box : Type := box (unbox : unit)`, or if `Box` is given primitive projections

VojtechStep:
Okay, this kind of makes sense. The term that coq-elpi generates with my patch is
```elpi
fun `x1` (global (indt «Box»)) c0 \
 fun `x2` (global (indt «Box»)) c1 \
  fun `x_R` (app [global (indt «Box_R»), c0, c1]) c2 \
    match c2
    (fun `x1` (global (indt «Box»)) c3 \
      fun `x2` (global (indt «Box»)) c4 \
        fun `x_R` (app [global (indt «Box_R»), c3, c4]) c5 \
        app
          [global (indt «unit_R»),
          match c3
            (fun `x` (global (indt «Box»)) c6 \
              global (indt «unit»))
            [fun `u` (global (indt «unit»)) c6 \ c6],
          match c4
            (fun `x` (global (indt «Box»)) c6 \
              global (indt «unit»))
            [fun `u` (global (indt «unit»)) c6 \ c6]])
    [fun `u1` (global (indt «unit»)) c3 \
      fun `u2` (global (indt «unit»)) c4 \
        fun `u_R` (app [global (indt «unit_R»), c3, c4]) c5 \ c5]
```
so equivalent to
```rocq
fun (x1 x2 : Box) (x_R : Box_R x1 x2) =>
  match x_R in (Box_R x3 x4)
        return (unit_R
                  match x3 with
                  | box u => u
                  end
                  match x4 with
                  | box u => u
                  end)
  with
  | box_R _ _ u_R => u_R
  end.
```

while what I should be generating is
```elpi
fun `x1` (global (indt «Box»)) c0 \
 fun `x2` (global (indt «Box»)) c1 \
  fun `x_R` 
   (app [global (indt «Box_R»), c0, c1]) 
   c2 \
   match c2 
    (fun `x_R` 
      (app [global (indt «Box_R»), c0, c1]) 
      c3 \
      app
       [global (indt «unit_R»), 
        app [global (const «unbox»), c0], 
        app [global (const «unbox»), c1]]) 
    [fun `u_R` 
      (app
        [global (indt «unit_R»), 
         app [global (const «unbox»), c0], 
         app [global (const «unbox»), c1]]) 
      c3 \ c3]
```
which is equivalent to
```rocq
fun (x1 x2 : Box) (x_R : Box_R x1 x2) =>
  match x_R in (Box_R _ _)
        return unit_R (unbox x1) (unbox x2)
  with
  | box_R _ _ u_R => u_R
  end.
```

Fair enough, records are no longer translated as indexed inductives, so `match` can't refine the indices. IMO pattern matching on records is evil, why are you not using inductives then, but sure, I'll need to find a way to support this.

PR #1042, opened 2026-06-12
Add builtin to the tc app

Tragicus:
Dependency of https://github.com/math-comp/hierarchy-builder/pull/596


apps/tc/src/rocq_elpi_class_tactics_takeover.ml
```diff
@@ -180,6 +180,16 @@ let set_solver_mode kind qname (l: Libnames.qualid list) =
   | ANone-> Lib.add_leaf (cache_solver_mode (qname, Set (Only OSet.empty), false))
   | ASet -> Lib.add_leaf (cache_solver_mode (qname, Set (Only l), false))
 
+let tc_builtins = [
+  Elpi.API.BuiltIn.MLCode(Pred("coq.TC.override-solver",
+    In(Elpi.API.BuiltInData.list Elpi.API.BuiltInData.string, "GR",
+    Easy("Declares that goals on GR should be solved using elpi's typeclass solver.")),
+  (fun gr ~depth ->
+    set_solver_mode AAdd ["TC.Solver"] (List.map Libnames.qualid_of_string gr))),
```
gares: I think that the tc solver should not be hardcoded

gares:
You should also add the new opam package to the ci job making the release, and add changelog entries

gares:
Davide we need a little doc for the filter api, in particular what happens if two tc problems are to be solved and a tc solver only handles one kind.

The use case of Quentin is that HB uses a very specific solver for classes it generates. It is really an ad hoc solver to implement some notations without hacking too much.

FissoreD:
IIRC when you have a cluster of typeclass goals, the custom solver is used only if it overrides all the class evars in it, otherwise, it uses rocq solver.

gares:
cluster means interdependent goals? like `?g1 : C1 ?t` + `?g2 : C2 ?t` or only if `?g1` occurs as argument of `C2`?

FissoreD:
I think the second.
The code is [here](https://github.com/rocq-prover/rocq/blob/1d8273d4977b07bf7252dda8fbc027c9d235bf8a/tactics/class_tactics.ml#L1147-L1164) and is used [here](https://github.com/rocq-prover/rocq/blob/1d8273d4977b07bf7252dda8fbc027c9d235bf8a/tactics/class_tactics.ml#L1250).
Each component of the cluster is passed to the first solver whose condition accepts it

gares:
So I think it makes sense for this specific application, all classes are generated by HB, hence solved by the same solver.

apps/tc/src/rocq_elpi_class_tactics_takeover.ml
```diff
@@ -180,6 +180,17 @@ let set_solver_mode kind qname (l: Libnames.qualid list) =
   | ANone-> Lib.add_leaf (cache_solver_mode (qname, Set (Only OSet.empty), false))
   | ASet -> Lib.add_leaf (cache_solver_mode (qname, Set (Only l), false))
 
+let tc_builtins = [
+  Elpi.API.BuiltIn.MLCode(Pred("coq.TC.override-solver",
+    In(Elpi.API.BuiltInData.string, "GR",
```
gares:
```suggestion
    In(Elpi.API.BuiltInData.string, "SolverName",
```

apps/tc/src/rocq_elpi_class_tactics_takeover.ml
```diff
@@ -180,6 +180,17 @@ let set_solver_mode kind qname (l: Libnames.qualid list) =
...
+    In(Elpi.API.BuiltInData.list Elpi.API.BuiltInData.string, "GR",
```
gares:
```suggestion
    In(Elpi.API.BuiltInData.list Elpi.API.BuiltInData.string, "ClassGR",
```

apps/tc/src/rocq_elpi_class_tactics_takeover.ml
```diff
@@ -180,6 +180,17 @@ let set_solver_mode kind qname (l: Libnames.qualid list) =
...
+    Easy("Declares that goals on GR should be solved using elpi's typeclass solver."))),
```
gares:
```suggestion
    Easy("Declares that instances of class ClassGR should be solved using elpi's typeclass solver SolverName."))),
```

FissoreD:
We had a discussion this morning with Yves, is it mandatory to have a doc folder in a plugin? 
It seems to me that the pattern used in those files is always the same. Can't it be generated?

dune-project
```diff
@@ -35,6 +35,12 @@
   ppx_optcomp
   (ocaml-lsp-server :with-dev-setup)))
 
+(package
+ (name rocq-elpi-typeclasses)
+ (synopsis "TC bindings for rocq-elpi")
+ (description "This package provides a typeclass solver written in Elpi for Rocq")
```
gares:
```suggestion
 (description "This package provides a typeclas solver written in Elpi, as well as the infrastructure for writing custom ones")
```

Also in the .opam file

dune-project
```diff
@@ -35,6 +35,12 @@
   ppx_optcomp
   (ocaml-lsp-server :with-dev-setup)))
 
+(package
+ (name rocq-elpi-typeclasses)
+ (synopsis "TC bindings for rocq-elpi")
```
gares:
```suggestion
 (synopsis "Typeclass solver in rocq-elpi")
```

gares:
> We had a discussion this morning with Yves, is it mandatory to have a doc folder in a plugin? It seems to me that the pattern used in those files is always the same. Can't it be generated?

Yes, I guess, but by which tool? It should be documented, that is for sure.

FissoreD:
> > We had a discussion this morning with Yves, is it mandatory to have a doc folder in a plugin? It seems to me that the pattern used in those files is always the same. Can't it be generated?
> 
> Yes, I guess, but by which tool? It should be documented, that is for sure.

Not sure how it works, but could we label the function declaring builtins with a custom `[@@ gen_doc]` which creates the few lines of code in gen_doc.ml and then we delagate to dune to build the .elpi with the new builtins?

apps/tc/src/rocq_elpi_class_tactics_takeover.ml
Tragicus: I kept the word `goal` because of the verb `solve`.

apps/tc/src/rocq_elpi_class_tactics_takeover.ml
Tragicus: Or even `ClassNames` since it is a list of strings.

PR #1046, opened 2026-06-17
Fix mfix bugs

Janno:


src/rocq_elpi_HOAS.ml
```diff
@@ -1724,12 +1724,15 @@ let rec constr2lp coq_ctx ~calldepth ~depth state t =
          let names = Array.to_list names in
          let tys = Array.to_list tys in
          let rargs = Array.to_list rargs in
-         let (n,state,env), names_rnos_tys =
-           CList.fold_left_map (fun (n,state,env) ((name,rno),typ0) ->
-             let state,typ = aux ~depth:(depth+n) env state typ0 in
-             let env = EConstr.push_rel Context.Rel.Declaration.(LocalAssum(name,typ0)) env in
-             (n+1,state,env), (name.Context.binder_name, rno, typ))
-           (0,state,env) (List.combine (List.combine names rargs) tys) in
+         let state, tys_lp = CList.fold_left_map (aux ~depth env) state tys in
```
gares: This looks wrong: tys are going to be placed under under lambdas, so they should live at larger depths.

You see this only if the type of a fix contains a binder, put a `(H : forall x:prop,x)` argument.

I guess there is another bug somewhere else.

gares: Oh, I see, the rocq ones all live at the current depth, but the elpi ones are shifter progressively. my bad

gares: nope, I war right. In the screenshot here it should be `c3\c3` and not `c3\c2`. So there must be another bug, for which this patch is not the right one.
Screenshot From 2026-06-17 15-21-24

gares (changes requested):

src/rocq_elpi_HOAS.ml
Janno: It might be fixed now. I have no idea what is going on so it is a little hard to tell. What's the right way to tests this? I don't think coq-elpi has Rocq output tests yet, does it?

gares:
My patch also has a second part:
```diff
commit b730f4f631a44a49077b537049854f27416be913 (HEAD -> janno/mfix-bugs)
Author: Enrico Tassi <Enrico.Tassi@Inria.fr>
Date:   Wed Jun 17 16:29:39 2026 +0200

    tentative fix

diff --git a/src/rocq_elpi_HOAS.ml b/src/rocq_elpi_HOAS.ml
index d2fb1713..52dd57eb 100644
--- a/src/rocq_elpi_HOAS.ml
+++ b/src/rocq_elpi_HOAS.ml
@@ -1724,15 +1724,13 @@ let rec constr2lp coq_ctx ~calldepth ~depth state t =
          let names = Array.to_list names in
          let tys = Array.to_list tys in
          let rargs = Array.to_list rargs in
-         let state, tys_lp = CList.fold_left_map (aux ~depth env) state tys in
-         let names_rnos_tys =
-           List.map2 (fun (name,rno) typ -> (name.Context.binder_name, rno, typ))
-             (List.combine names rargs) tys_lp in
-         let env =
-           List.fold_left (fun env (name,typ0) ->
-             EConstr.push_rel Context.Rel.Declaration.(LocalAssum(name,typ0)) env)
-             env (List.combine names tys) in
-         let n = List.length names in
+         let (n,state,env), names_rnos_tys =
+           CList.fold_left_map (fun (n,state,env) ((name,rno),typ0) ->
+             let state,typ = aux ~depth env state typ0 in
+             let typ = U.move ~from:depth ~to_:(depth+n) typ in
+             let env = EConstr.push_rel Context.Rel.Declaration.(LocalAssum(name,typ0)) env in
+             (n+1,state,env), (name.Context.binder_name, rno, typ))
+           (0,state,env) (List.combine (List.combine names rargs) tys) in
          let state, bos = CArray.fold_left_map (aux ~depth:(depth+n) env) state bos in
          state, in_elpi_mfix names_rnos_tys focus_idx (Array.to_list bos)
     | C.Proj(p,_,t) ->
@@ -2355,14 +2353,16 @@ and lp2constr ~calldepth syntactic_constraints coq_ctx ~depth state ?(on_ty=fals
       let outer_ctx = coq_ctx in
       let focus_idx = lp2int ~depth ~ctx:"mfix focus" focus_lp in
       let top_rno = lp2int ~depth ~ctx:"mfix rno" rno in
+      let mfix_depth = depth in
       let rec collect_ty ~depth state coq_ctx node defs gls_acc =
         match E.look ~depth node with
         | E.App(c2,name_lp,[rno_lp; ty_lp; rest_lam]) when mfix_tyc == c2 ->
           let name = in_coq_fresh_annot_name ~depth ~coq_ctx depth name_lp in
           let rno = lp2int ~depth ~ctx:"mfix-ty rno" rno_lp in
           let state, ty, gl =
+            let ty_lp = U.move ~from:depth ~to_:mfix_depth ty_lp in
             lp2constr ~calldepth syntactic_constraints outer_ctx
-              ~depth state ~on_ty:true ty_lp in
+              ~depth:mfix_depth state ~on_ty:true ty_lp in
           let coq_ctx = push_coq_ctx_local depth
             (Context.Rel.Declaration.LocalAssum(name,ty)) coq_ctx in
           let defs = (name, rno, ty) :: defs in
```

gares:
I think the two bugs were compensating eachothers in some cases.

Your code for the first part is OK, btw, no need to use mine (also the env update was fishy, but I could not find a way to see a bug with it)

Janno:
I added the second half of your commit.

EDIT: I added too many commits.

EDIT 2: Should be fixed.

gares:
I'm a bit lost, do you want me to push a test for needed to show that 
```diff
@@ -2355,14 +2353,16 @@ and lp2constr ~calldepth syntactic_constraints coq_ctx ~depth state ?(on_ty=fals
       let outer_ctx = coq_ctx in
       let focus_idx = lp2int ~depth ~ctx:"mfix focus" focus_lp in
       let top_rno = lp2int ~depth ~ctx:"mfix rno" rno in
+      let mfix_depth = depth in
       let rec collect_ty ~depth state coq_ctx node defs gls_acc =
         match E.look ~depth node with
         | E.App(c2,name_lp,[rno_lp; ty_lp; rest_lam]) when mfix_tyc == c2 ->
           let name = in_coq_fresh_annot_name ~depth ~coq_ctx depth name_lp in
           let rno = lp2int ~depth ~ctx:"mfix-ty rno" rno_lp in
           let state, ty, gl =
+            let ty_lp = U.move ~from:depth ~to_:mfix_depth ty_lp in
             lp2constr ~calldepth syntactic_constraints outer_ctx
-              ~depth state ~on_ty:true ty_lp in
+              ~depth:mfix_depth state ~on_ty:true ty_lp in
           let coq_ctx = push_coq_ctx_local depth
             (Context.Rel.Declaration.LocalAssum(name,ty)) coq_ctx in
           let defs = (name, rno, ty) :: defs in
```
is needed?

gares:
I will, but later

Janno:
Oh, maybe those changes were not necessary at all? I am also confused.

gares:
I think they are, but I dont have a test. Let me push it and we will have proof.

gares:
No, I cannot break it.

The safeguard is that it uses outern_ctx, hence it already performs the "move" and also checks you don't use any fixpoint name in the type of the other ones.

With my `move` it is still correct, but the error message is way more ugly (pruning failure in HO unif).

So please revert my very last hunk. The test is
```
Fail Elpi Query lp:{{

  T = (fun `x` {{nat}} c0 \
 mfix 0 0 
  (mfix-ty `_elpi_renamed_x_1` 0 
    (prod `n` {{nat}} c1 \
      prod `H` (prod `P` (sort prop) c2 \ c2) c2 \ {{nat}}) 
    c1 \
    mfix-ty `y` 0 
     (prod `n` {{nat}} c2 \
       prod `H` (prod `P` (sort prop) c3 \ c1 /*bad */) c3 \ {{nat}}) 
     c2 \
     mfix-bo
      [(fun `n` {{nat}} c3 \
        fun `H` (prod `P` (sort prop) c4 \ c4) c4 \
         match c3 
          (fun `_elpi_renamed_n_5` {{nat}} c5 \
            X0 c0 c1 c2 c3 c4 c5) 
          [{{O}}, 
           fun `n'` (X1 c0 c1 c2 c3 c4) c5 \ app [c2, c5, c4]]), 
       fun `n` {{nat}} c3 \
        fun `H` (prod `P` (sort prop) c4 \ c4) c4 \
         match c3 
          (fun `_elpi_renamed_n_5` {{nat}} c5 \
            X2 c0 c1 c2 c3 c4 c5) 
          [app [{{S}}, {{O}}], 
           fun `n'` (X3 c0 c1 c2 c3 c4) c5 \ app [c1, c5, c4]]])),
  coq.say T,
  std.assert-ok! (coq.typecheck T _) "illtyped"


}}.
```

gares (approved):

gares:
thanks, if you could rebase the mutind branch on top of the new master it would be great.

Janno:
Thanks for taking over. I managed to force-push away one of my changes but I think you included your version instead so I think it all worked out.

Janno:
I opened https://github.com/LPCIC/coq-elpi/pull/1053 with a cleaner set of commits without build system changes. I kept all the other commits for now but most of them should probably be squashed.

PR #1053, opened 2026-06-19
Support for mutual inductives

Janno:


gares:
Thanks, I started to review some code but the current structure is hard to review/merge.

I think it would help to have all changes about .ml files and files in elpi/ and tests/ to be part of a first PR.

Then have all changes about derive/eq to be in a followup commit in another PR.

And so on. This way I can merge things incrementally, and add tests on each of these.

Can you squash then separate the changes in this way?

VojtechStep:
Btw is this the same code as https://github.com/SkyLabsAI/rocq-elpi/pull/2?

Janno:
> Btw is this the same code as [SkyLabsAI#2](https://github.com/SkyLabsAI/rocq-elpi/pull/2)?

Yes, this is the same code rebased on master (instead of our own vendored snapshot).

> Can you squash then separate the changes in this way?

Yes, I'll try to get that done ASAP.

Janno:
Superseded by #1054 and #1056

PR #1054, opened 2026-06-19
Add HOAS support for mutual inductive types

Janno:
This is #1053 restricted to `src/`, `elpi/`, and `test/`. It should contain only changes related to the HOAS representation of mutual inductive types.

- [x] `coq.env.indt-block` should also also output the knames and types as `coq.env.indt` does

### DIFF
--- a/src/rocq_elpi_HOAS.ml
+++ b/src/rocq_elpi_HOAS.ml
@@ -3748,7 +3748,7 @@ type record_decl = {
 type ind_decl =
   | Inductive of inductive_decl
   | Record of record_decl
-  | Mutual of ind_decl list
+  | Mutual of inductive_decl list
 type hoas_ind = {
   params : Glob_term.binding_kind ctx_entry list;
   decl : ind_decl;
@@ -3867,10 +3867,7 @@ let hoas_ind2lp ~depth coq_ctx state { params; decl } =
       let state, ks, gls2 =
         API.Utils.map_acc embed_constructor state constructors in
       state, in_elpi_indtdecl_inductive state kind (Name id) arity ks, List.flatten [gls1 ; gls2]
-   | Mutual decls ->
-      let inds = List.map (function
-        | Inductive i -> i
-        | Record _ | Mutual _ -> nYI "records inside mutual inductive blocks") decls in
+   | Mutual inds ->
       let ninds = List.length inds in
       if ninds < 2 then nYI "ill-formed mutual inductive block";
       let sigma = get_sigma state in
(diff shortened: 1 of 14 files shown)


gares:
Failures are spurious, it is Rocq failing, not the Elpi code: https://github.com/rocq-prover/rocq/issues/22149

gares:
The fact that the whole CI passes, and that I replaced my old code for 1 inductive with the new code for n inductives, makes this PR pretty ready IMO.

Janno:
As soon as you declare this done and it is merged I can work on (or rather have the LLMs work on) rebasing the remaining changes

src/rocq_elpi_HOAS.ml
```diff
@@ -3492,6 +3508,111 @@ let lp2inductive_entry ~depth coq_ctx constraints state t =
...
+  let aux_mutual_constructors coq_ctx ~depth (params,impls) inds state constructor_blocks =
+    let params = force_name_ctx params in
+    let paramsno = List.length params in
+    let ninds = List.length inds in
+    if List.length constructor_blocks <> ninds then
+      err Pp.(str"mblock: expected " ++ int ninds ++ str" constructor blocks, got " ++
+              int (List.length constructor_blocks));
+    let state, constructor_infos, gls_rev =
+      List.fold_left2 (fun (state, infos, gls_acc) (name, nuparams, nuimpls, arity, _kind) ks ->
+        if nuparams <> [] then nYI "non-uniform parameters in mutual inductives";
```
Janno, line 3538: Is this a regression compared to non-mutual inductives? I haven't been keeping up with the uniform parameter changes.

src/rocq_elpi_HOAS.mli
Janno, line 1: Do you want to expose similar helpers to the ones exposed for non-mutual inductives? Those would be `in_elpi_indtdecl_minductive` and `in_elpi_indtdecl_mblock`, I think

src/rocq_elpi_HOAS.ml
gares: Isn't this only called for mutual inductives?

src/rocq_elpi_HOAS.mli
gares: I'm all for cleaning up code, but I don't see exactly what you mean.
Were would they be used?

src/rocq_elpi_HOAS.ml
Janno: Sorry, "regression" was perhaps the wrong word. It might be a missing feature compared to non-mutual inductives.

src/rocq_elpi_HOAS.mli
Janno: I am not sure, hence the question. I guess they weren't needed for the derive changes so maybe that's fine. It just seemed odd to have some functions exposed for non-mutual inductives but not for mutual inductives.

gares: The one for regular inductives is use in the argumnets HOAS and glob, i.e. what is needed when you write `derive Inductive`. I guess these are still nYI code paths then.

src/rocq_elpi_HOAS.ml
gares: yes, I don't mind having it added later. Maybe a test triggering it (if not already present) would be nice

src/rocq_elpi_HOAS.mli
Janno: I think all the tests in the derive changes use `Inductive bla := .. . #[only(..)] derive bla.`. Would that explain it?

gares: yes, and it is perfectly fine as a first step.

gares (approved):

Issue #871, opened 2025-08-29
Bad interaction between `apply:` and typeclass solver


Tragicus:
It seems that `apply:` calls the typeclass solver on the goal it produces and somehow accepts a partial solution from elpi's typeclass solver. I am surprised both by the fact that the typeclass solver gives a partial answer and that `apply:` accepts it.
```coq
From Corelib Require Import ssreflect.

Class A := {}.

Goal (nat -> A) -> A.
Proof.
move=> hyp.
apply: id.
(* Goal : A *)
Abort.

From elpi Require Import tc.

Class B := {}.

Goal (nat -> B) -> B.
Proof.
move=> hyp.
apply: id.
(* Goal : nat *)
Abort.
```
Also, if I replace `B` by `A` in the last `Goal`, the `apply: id` command fails, which is even weirder.
```coq
Goal (nat -> A) -> A.
Proof.
move=> hyp.
Fail apply: id.
```

gares:
Would you be so nice to attach the output of Set Debug Typeclasses for the first run? I suspect it runs, solves it using hyp, but then we toss the result as inconclusive. Maybe because the nat goal is not shelved.

And there is clearly some bug...

PR #1055, opened 2026-06-21
Check existence of TC predicate before make class queries/rules

FissoreD:


FissoreD:
this is partially linked with #871.
```
Goal (nat -> A) -> A.
Proof.
move=> hyp.
Fail apply: id.
```
In the given example, the goal fails since the predicate for the calss A is not declared in Elpi.
This explains the different behavior with the goal talking about the class B.

PR #1059, opened 2026-07-01
release 3.5.0

gares:


gares:
@CohenCyril let's look at CI

CohenCyril:
@affeldt-aist
Does that ring a bell?

 https://github.com/LPCIC/coq-elpi/actions/runs/28522775050/job/84556050375#step:13:149

> File "./realsum.v", line 14, characters 29-33:
> Error: Cannot find a physical path bound to logical path
> esum with prefix mathcomp.

affeldt-aist:
Maybe related: the experimental package which contains realsum.v now depends on the analysis package (before that it was depending on classical and reals only)

CohenCyril:
There is a bug that resists my investigations here. I will try harder on Monday

PR #1065, opened 2026-07-08
Binary parametricity translation of primitive records

VojtechStep:
Take 2.
Parts of the previous PR (#1027) ended up being reverted, because one can pattern match on records without primitive projections, and changing the translation of records meant that I had to change the translation of match motives as well, which I hadn't done.

What I propose here is keeping the translation of records without primitive projections as indexed inductives (since for all intents and purposes they are one-constructor inductives), but translate records with primitive projections as records with primitive projections.

The translation of non-primitive records is implemented on top of the HOAS translation of inductives, so the older implementation on top of `coq.build-indt-decl` has been removed.

I also added a check that the user doesn't try to manually translate projections of primitive records, since this is something that needs to be done for non-primitive records, but doesn't make sense for primitive projections, similarly to inductive constructors. AFAICT the explicit check is necessary, because when the user tries to translate a projection, the term that elpi receives is the compatibility constant, which is just a const, not a primitive, and constants are allowed to be translated multiple times (I don't know why, but that's orthogonal)

### DIFF
--- a/apps/derive/elpi/param2.elpi
+++ b/apps/derive/elpi/param2.elpi
@@ -228,6 +228,8 @@ param-indt-decl I Suffix NR DeclR :-
   % We can avoid the issue by first converting all parameters to non-uniform
   % ones, since then passing the unapplied inductives everywhere still produces
   % well-typed arities.
+  % As usual, we treat record types as inductives with a single constructor;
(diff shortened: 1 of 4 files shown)


gares (approved):

gares:
thanks

PR #1067, opened 2026-07-10
declare goals as future goals (avoid accidental TC resolution)

gares:
Apparently some tactics, but not all, trigger a TC resolution of all evars that are TC and are not flagged as goals.

CC @ybertot 
CC @SkySkimmer I could not really find the link between future goals and TC resolution. I'd like a confirmation I'm not just moving the problem.

### DIFF
--- a/src/rocq_elpi_HOAS.ml
+++ b/src/rocq_elpi_HOAS.ml
@@ -3013,6 +3013,7 @@ let tclSOLUTION2EVD ~eta_contract_solution sigma0 solution gls =
     let roots = List.fold_right Evar.Set.add gls Evar.Set.empty in
     let sigma, declared_goals, shelved_goals = solution2evd ~eta_contract_solution sigma0 solution roots in
     let sigma = Evd.fold_future_goals Evd.remove_future_goal sigma in
+    let sigma = List.fold_right Evd.declare_future_goal declared_goals sigma in
     (* debug Pp.(fun () -> str "Old Shelved Goals: " ++ prlist_with_sep spc Evar.print sh);
     debug Pp.(fun () -> str "New Evd Shelved Goals: " ++ prlist_with_sep spc Evar.print (Evd.shelf sigma)); *)
   tclTHENLIST [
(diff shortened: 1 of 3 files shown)


src/rocq_elpi_HOAS.ml
```diff
@@ -3013,6 +3013,7 @@ let tclSOLUTION2EVD ~eta_contract_solution sigma0 solution gls =
     let roots = List.fold_right Evar.Set.add gls Evar.Set.empty in
     let sigma, declared_goals, shelved_goals = solution2evd ~eta_contract_solution sigma0 solution roots in
     let sigma = Evd.fold_future_goals Evd.remove_future_goal sigma in
+    let sigma = List.fold_right Evd.declare_future_goal declared_goals sigma in
```
SkySkimmer: does the remove_future_goal line still do something with this? if so what is it for?

SkySkimmer: also doing declare_future_goal just before you tclSETGOALS the same evars seems strange

gares: I don't know, I can try to remove it.

What I really would like is that this tactic, among the new evars, flags correctly declared_goals and shelved_goals. The sigma may be passed to ltac code, so I'm not so sure the list of future goals in empty in there.

Even without elpi in the picture, I don't know who is responsible for this data, which component owns it.

SkySkimmer: future goals are evars which will end up as goals or in the shelf depending on the surrounding tactic 
in particular refine makes them goals, so maybe you want to do like refine:
~~~ocaml
  (* avoid touching unrelated future goals *)
  let sigma = Evd.push_future_goals sigma in
  tclEVARS sigma <*>
  run whatever tactic >>= fun tac_results ->
  tclEVARMAP >>= fun sigma ->
  let future_goals, sigma = Evd.pop_future_goals sigma in
  let future_goals = Evd.FutureGoals.map_filter (Proofview.Unsafe.advance sigma) future_goals in
  (* not sure the below filter is actually needed *)
  let future_goals = Evd.FutureGoals.filter (fun ev -> not @@ Evd.mem_shelf ev sigma) future_goals in
  let sigma = (* here refine defines the focused goal with the body generated by the tactic *) in
  let sigma = Proofview.Unsafe.mark_as_goals sigma (Evd.FutureGoals.comb future_goals) in
  let comb = CList.rev_map (fun x -> Proofview.goal_with_state x state) (Evd.FutureGoals.comb future_goals) in
  Proofview.Unsafe.tclEVARS sigma <*>
  Proofview.Unsafe.tclSETGOALS comb
~~~

PR #1072, opened 2026-07-21
Avoid global ref in qualified_name parsing

SkySkimmer:


SkySkimmer:
github changed something so nix CI is broken
docker CI uses an older checkout action version so is fine

PR #1078, opened 2026-07-24
[CI] Update Nix and add Rocq 9.3

proux01:


gares:
Thanks!

Is there a way to remove these pr_target jobs? They now fails with:

Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.

proux01:
> Is there a way to remove these pr_target jobs? They now fails with:

Not that I know, we still need to ignore them when the corresponding action file is updated.
The new error will be fixed after merging this one.

proux01:
@gares CI "green" (the jasmin failure is known and unrelated and the pull_request_target failures are to be ignored)

PR #1092, opened 2026-08-10
Fix inductive parameter relevance; new coq.string->name-irrelevant and coq.typecheck-relevance

JasonGross:
Elpi code that constructs binders over SProp types can leave relevant marks on irrelevant binders, causing the kernel error:

```
Binder ... has relevance mark set to Relevant but was expected to be
Irrelevant (maybe a bugged tactic)
```

This adds `coq.string->name-irrelevant` alongside `coq.string->name-relevant` for constructing an irrelevant name, and `coq.typecheck-relevance T T1` alongside `coq.typecheck-ty` for recomputing every binder mark from its type. If a binder's type does not retype, its existing mark is retained.

The latter uses `Retyping.relevance_of_type` and catches only `Retyping.RetypeError`. It fixes the marks a node's own binders carry and leaves the descent to `Termops.map_constr_with_full_binders`, so a binder under a `Case`, `Fix`, or `CoFix` is retyped with the enclosing binders in scope. Without that context, an open branch body can silently derive marks from unrelated ambient de Bruijn indices. The final two tests in `tests/test_HOAS.v` exercise cases where the answers differ; `coq.env.add-const` confirms that the kernel accepts the corrected marks.

#1093 remains separate because its readback change is an independent behavior question.

Wordsmithed by Codex.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,12 @@
   `Invalid_argument("index out of bounds")` from the kernel, printing one
   was producing an ill-typed declaration)
 
+### API:
+- New `coq.string->name-irrelevant`, the counterpart of
+  `coq.string->name-relevant`
+- New `coq.typecheck-relevance` recomputing, by retyping, the relevance mark of
+  every binder of a term
+
 ### LIB:
 - New  `coq.indt-decl->ids`, `coq.mutual?`, `coq.mutual.members`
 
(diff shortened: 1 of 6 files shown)


gares (changes requested):
There is already `coq.name.relevant?` to test, and `coq.string->name-relevant` (in addition to `coq.string->name`) to build a relevant name.

The other API is missing, but the name is odd, maybe `coq.typecheck-relevance` or something like that. I don't think there is any `coq.term.*` API around.

gares:
Also `coq.string->name-irrelevant` would be ok.

JasonGross:
Should be updated now

gares (approved):
Thanks.

PR #1094, opened 2026-08-10
Fix minductive when a member arity mentions a parameter

JasonGross:
`coq.env.add-indt` raises

```
Anomaly "Uncaught exception Invalid_argument("index out of bounds")."
```

for any mutual inductive block in which the arity of a member mentions a parameter of the block.

### Reproducer

```coq
Elpi Command tB.
Elpi Accumulate lp:{{
main _ :-
  D = (parameter "A" explicit {{ Type }} a\
    (minductive "stree" tt (arity {{ Type }}) t\
    (minductive "sforest" tt (arity {{ lp:a -> Type }}) f\
      (mblock [
        [constructor "snode" (parameter "x" explicit a x\ arity {{ lp:f lp:x -> lp:t }})],
        [constructor "sempty" (parameter "x" explicit a x\ arity {{ lp:f lp:x }})]
      ])))),
  std.assert-ok! (coq.typecheck-indt-decl D) "illtyped",
  coq.env.add-indt D _.
}}.
Elpi tB.
```

`coq.typecheck-indt-decl` and `coq.elaborate-indt-decl-skeleton` both accept the declaration; the failure only shows up in the kernel, at `IndTyping.check_arity` ← `typecheck_inductive` ← `Safe_typing.add_mind`.

Shape of the bug:

- one `minductive` with a parameter-dependent arity: works
- two members with parameter-*in*dependent arities: works (this is what `tests/test_mutind.v` covered, hence the miss)
- two members where any arity mentions a parameter: anomaly

### Cause

`aux_mind_lam` extends the `coq_ctx` with one binder per member as it walks the chain of `minductive` nodes, so `aux_inductive` reads the arity of the *i*-th member in a context that already holds the binders of the *i* preceding ones: its parameters sit at de Bruijn `i+1 .. i+paramsno`. `aux_mutual_constructors` then closes that arity with `it_mkProd_or_LetIn arity params`, which expects the parameters at `1 .. paramsno`. Constructor types were already relocated for exactly this reason (`relocate_mutual_constructor_type`); arities were not.

While writing the regression test I found the mirror image of the same defect in the printing direction. `hoas_ind2lp`'s `embed_arities` pushes a binder per member into `coq_ctx` but embeds each arity unlifted, so for

```coq
Inductive stree (A : Type) : A -> Type :=
| snode (x : A) (f : sforest A x) : stree A x
with sforest (A : Type) : A -> Type :=
| sempty (x : A) : sforest A x
| scons (x : A) (t : stree A x) (f : sforest A x) : sforest A x.
```

`coq.env.indt-decl` used to produce

```
parameter A explicit (sort (typ «stree.u0»)) c0 \
 minductive stree tt (arity (prod `_` c0 c1 \ sort (typ ...))) c1 \
  minductive sforest tt (arity (prod `_` c1 c2 \ sort (typ ...))) c2 \
   ...
```

where `sforest`'s arity is indexed by `c1`, the `stree` binder, instead of `c0`, the parameter. The resulting `indt-decl` is ill-typed, so `coq.env.indt-decl` followed by `coq.typecheck-indt-decl` fails on such a block.

### Fix

Relocate the arity of member *i* across the *i* preceding binders, in both directions. An arity may not mention the inductive types being defined; that is now reported as an error rather than silently reinterpreted as a parameter.

Non-mutual inductives take the same code path with `i = 0` and are unaffected.

### Tests

`tests/test_mutind.v` gains a block whose members are both indexed by a parameter, exercising the two directions (`coq.env.indt-decl` on a Rocq-declared block, and `coq.env.add-indt` on an elpi-built one) plus the existing `Module Type` equivalence check.

Verified against Rocq `master`: without the patch, the `coq.env.add-indt` reproducer above raises the anomaly and the printing half produces an ill-typed declaration; with it both succeed, and `dune build tests` + `dune runtest tests` pass.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,12 @@
 
 # UNRELEASED
 
+### HOAS:
+- Fix `minductive` when the arity of a member of the block mentions a
+  parameter, in both directions (reading a declaration back was raising
+  `Invalid_argument("index out of bounds")` from the kernel, printing one
+  was producing an ill-typed declaration)
+
 ### LIB:
 - New  `coq.indt-decl->ids`, `coq.mutual?`, `coq.mutual.members`
 
--- a/src/rocq_elpi_HOAS.ml
+++ b/src/rocq_elpi_HOAS.ml
@@ -3566,6 +3566,20 @@ let lp2inductive_entry ~depth coq_ctx constraints state t =
     EC.Vars.substl subst t
   in
 
+  (* The arity of the i-th member of a mutual block is read back in a context
+     already extended with the binders of the i preceding members (see
+     aux_mind_lam), hence its parameters live at de Bruijn i+1..i+paramsno.
+     Arities cannot mention the inductive types being defined, so we drop these
+     i binders and bring the parameters back to 1..paramsno, which is what
+     it_mkProd_or_LetIn expects below. *)
+  let relocate_mutual_arity sigma ~name ~nindsbefore t =
+    if nindsbefore = 0 then t
+    else if not (EConstr.Vars.noccur_between sigma 1 nindsbefore t) then
+      err Pp.(str"minductive: the arity of " ++ Id.print name ++
+              str" mentions an inductive type of the same block")
+    else EConstr.Vars.liftn (-nindsbefore) (nindsbefore + 1) t
+  in
+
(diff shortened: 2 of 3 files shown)


gares (approved):

gares:
Thanks

PR #1095, opened 2026-08-11
Build binders from the relevance of their type, not of their id (negative result)

JasonGross:
**Negative result — this does not work. Keeping it open only as the record of
the experiment; #1093 is the fix.** CI fails on exactly the test it was meant
to make pass, on every Rocq version:

```
File "./tests/test_HOAS.v", line 1029:
Error: Binder (proof : "sTrue") has relevance mark set to relevant but was
expected to be irrelevant (maybe a bugged tactic).
Raised at Type_errors.error_bad_binder_relevance
Called from Typeops.check_assum_annot
Called from Typeops.check_context.(fun)
Called from IndTyping.typecheck_inductive
```

That backtrace is the point: the mark the kernel rejects is on the *parameter
context of the entry*, which `lp2inductive_entry` builds by reading the `id`
back, not one of the marks the elpi helpers patched here. Fixing
`coq.arity->term` and `@pi-parameter` changes the context the declaration is
elaborated in, and changes nothing about the entry that reaches the kernel.
So there is no fix for this at the `coq.id->name` layer; it has to be at
readback, which is what #1093 does.

The observation below still stands on its own — `coq.arity->term` does produce
an ill-marked `prod` for an SProp parameter — it is just not what breaks
`coq.env.add-indt`.

---

<details>
<summary>Original description</summary>

An `indt-decl` carries its parameter names as `id`s, which have no relevance
mark of their own. Two elpi helpers invent one:

```elpi
coq.arity->term (parameter ID _ Ty Rest) (prod Name Ty R) :-
  coq.id->name ID Name, ...

macro @pi-parameter ID T F :-
  sigma N\ (coq.id->name ID N, pi x\ decl x N T => F x).
```

Since 3b62afd2 `coq.id->name` is `coq.string->name-relevant`, so both answer a
*relevant* binder whatever the type is:

```coq
Inductive sTrue : SProp := sI.

Elpi Query lp:{{
  coq.locate "sTrue" (indt ST),
  coq.arity->term
    (parameter "proof" explicit (global (indt ST)) _\ arity {{ Type }}) T,
  T = prod N _ _, coq.name.relevant? N R.   % R = some tt
}}.
```

and the kernel rejects such a term on the mark alone.

Reverting the hack does not help either: it puts a fresh relevance *variable*
back on the name, nothing solves it on this path, and it is defaulted to
relevant. Measured on a build that has 3b62afd2:

| name for `proof : sTrue` | `coq.env.add-const` |
|---|---|
| `coq.id->name` (relevant) | rejected |
| `coq.string->name` (relevance variable) | rejected |
| mark taken from the type | accepted |

So this adds `coq.id->name-of-ty ID Ty Name`, `coq.id->name` with the mark of
`Ty`, and uses it in `coq.arity->term`, `@pi-parameter` and `@pi-inductive`.
`coq.id->name` itself is untouched.

</details>

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,10 @@
   parameter, in both directions (reading a declaration back was raising
   `Invalid_argument("index out of bounds")` from the kernel, printing one
   was producing an ill-typed declaration)
+- New `coq.id->name-of-ty`, building a name whose relevance mark is the one of
+  the given type; `coq.arity->term`, `@pi-parameter` and `@pi-inductive` use it
+  instead of `coq.id->name`, so a parameter whose type lives in `SProp` no
+  longer gets a relevant binder that the kernel rejects
 
 ### API:
 - New `coq.string->name-irrelevant`, the counterpart of
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,12 @@
   `Invalid_argument("index out of bounds")` from the kernel, printing one
   was producing an ill-typed declaration)
 
+### API:
+- New `coq.string->name-irrelevant`, the counterpart of
+  `coq.string->name-relevant`
+- New `coq.typecheck-relevance` recomputing, by retyping, the relevance mark of
+  every binder of a term
+
 ### LIB:
 - New  `coq.indt-decl->ids`, `coq.mutual?`, `coq.mutual.members`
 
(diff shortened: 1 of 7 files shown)


JasonGross:
Closing: this does not work. CI here fails on `tests/test_HOAS.v` with the original `error_bad_binder_relevance`, from `IndTyping.typecheck_inductive` via `Typeops.check_context` — the rejected mark is on the parameter context of the entry, built during readback in `lp2inductive_entry`, not by `coq.arity->term` or `@pi-parameter`. So the `coq.id->name` layer cannot fix it and #1093 stands. Details in the description, kept for the record.

PR #1100, opened 2026-09-01
Backport to 9.3 of https://github.com/rocq-prover/rocq/pull/22161

proux01:


proux01:
This was a bad idea, let's not do it.

PR #1103, opened 2026-09-15
Avoid contaminating clients with `Set Uniform Inductive Parameters`

### LINKED ISSUE #1034: Avoid contaminating clients with `Set Uniform Inductive Parameters`
Since https://github.com/LPCIC/coq-elpi/commit/ec90c712ddc60d28cc7419c29f43b4bcc16be9ed (authored by @proux01), importing `derive` sets `Uniform Inductive Parameters` with `export` locality, which breaks client code. This causes problems for us, and is likely gonna affect others.  I think it would make sense to revert that change.

CC @Janno.

rlepigre-skylabs-ai:
Fixes #1034.


builtin-doc/elpi-builtin.elpi
VojtechStep, line 1: You seem to be using an older version of elpi, which regenerates the elpi-builtin file to its state before 3.7.2

PR #1106, opened 2026-09-18
New syntax

gares:


gares:
@proux01 have a look at the new deps in the opam file when you fix nix.
```
       > 532 |  (requires atdgen-runtime)
       >                  ^^^^^^^^^^^^^^
       > Error: Library "atdgen-runtime" not found.
```

proux01:
Pretty impressed by the LLM generated hack (ugly but should work)

gares:
I don't even understand it, but seems to work, so you can merge the PR on rocq now, if you like

gares:
> Pretty impressed by the LLM generated hack (ugly but should work)

I copy here the first answer the LLM gave me, for reference
```
Fix: added "atdgen-runtime" to rocq-elpi.opam's depends 
(this is what the Nix job actually reads), and mirrored it in dune-project
which is the documented source of truth for the opam file, also fixing the elpi version bound 
there which was stale (3.0..3.2.0 vs. the real 3.8.0..3.9.0). No source code changes needed — 
this is purely a missing build dependency declaration.
```

I guess now you are a bit less impressed ;-)

proux01:
> you can merge the PR on rocq now, if you like

Done, the overlay can be merged.

PR #1107, opened 2026-09-18
Merge pglobal and global constructors

gares:


gares:
@mattam82 this PR breaks all elpi clients, including trocq and trakt.

I still think this is a good change in the long run, but it is a bit unclear if we want to inflict this pain before the upoly branch is finalized

mattam82:
The idea was to push this first as it should introduce no change of behavior, but indeed changes the API. Later the upoly changes can come, which should be much less intrusive on the `.elpi` side for clients.

